PR #78, opened 2021-11-05
Adding the support for literal strings (without multiline strings)

akuleshov7:


akuleshov7:
Will fix diktat issues in the next PR

Issue #49, opened 2021-06-22
General feedback on the API design


Olivki:
Hello, this isn't regarding one specific issue per se, but rather some general feedback regarding the design of the current API. If any of this comes off as aggressive/mean sounding, I apologize, my intention is solely for constructive criticism.

Most of my opinions will be based on the API design of officially supported format libraries developed by JetBrains themselves, which can be found [here](https://github.com/Kotlin/kotlinx.serialization/tree/master/formats), and the Kotlin coding conventions provided by JetBrains, which can be found [here](https://kotlinlang.org/docs/coding-conventions.html). I'm not sure how much stuff you wanna change, but I figured it would be best to provide feedback while the library is still in early development, as a lot of these changes would break backwards compatibility.

There's a decent chunk of stuff that I want to provide feedback on, so I'm sorry if things read like a jumbled mess. I will try to section off the feedback to their own "sections" as best as I can.

----

If any of the suggestions here are something you like, I can make a pull requests with the fixes if desired. I would rather just explain my reasoning and thoughts before just making a pull requests with all fixes.

----

### The class name
First point to address here is the name, if we look at [naming rules](https://kotlinlang.org/docs/coding-conventions.html#naming-rules), it states that `Names of classes and objects start with an uppercase letter and use camel case`, and with camel case, each new *word* should be capitalized, and for acronyms, each letter representing the word should normally be treated as a new word.  meaning that if we follow these rules, the appropriate name for the class should be `KToml` rather than `Ktoml` as the `K` stands for Kotlin, and `toml` should be treated as one word. 
(By following the above rules it should technically be `KTOML`, but if we look at the officially supported formats like [json](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/json/commonMain/src/kotlinx/serialization/json/Json.kt#L54), and other classes developed by JetBrains, they seem to follow the rules of Dart wherein an acronym that's 3 or more characters long should be treated as a word, so instead of `URL` it would be `Url`.)

However, if we look at essentially all other libraries, even those outside of the officially supported formats, like [yamlkt](https://github.com/Him188/yamlkt/blob/master/yamlkt/src/commonMain/kotlin/net.mamoe.yamlkt/Yaml.kt) and [avro4k](https://github.com/avro-kotlin/avro4k/blob/master/src/main/kotlin/com/github/avrokotlin/avro4k/Avro.kt#L121), they just use the format name as the class name, meaning that rather than `Ktoml` it would be `Toml`.

[...]

PR #91, opened 2022-01-11
Tweak config and exception

Peanuuutz:
[here](https://github.com/akuleshov7/ktoml/issues/49#issuecomment-1009497563)


ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/exceptions/TomlSerializationExceptions.kt
```diff
@@ -0,0 +1,38 @@
...
+ * Internal exceptions used in the project
+ */
+
+@file:Suppress("MISSING_KDOC_TOP_LEVEL")
+
+package com.akuleshov7.ktoml.exceptions
+
+import kotlinx.serialization.SerializationException
+
+public sealed class TomlDecodingException(message: String) : SerializationException(message)
```
akuleshov7, line 11: If you have finally renamed it - may be make all exception's naming consistent? 
Let's select one of two variants:
1) All of them are with 'Toml' prefix
2) All of them are without a `Toml` prefix and only the base class `TomlDecodingException` contains it

akuleshov7:
Looks good, but it will completely break all users in the community 😭 😭 😭

Peanuuutz:
> Looks good, but it will completely break all users in the community 😭 😭 😭 

I have that in mind, so this problem could be solved later. Just put an @Deprecated to inform users about this change. I'll rollback soon and commit a new one like I said.

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/exceptions/TomlSerializationExceptions.kt
Peanuuutz: > All of them are without a `Toml` prefix and only the base class `TomlDecodingException` contains it

I prefer this, and I imagine something like `TomlEncodingException` for encoding. I'd like to make both of these as `public` for users to manipulate, while actual exception types keep `internal`.

Peanuuutz: A further question: If there is only one file in exception folder, I think to put it out along with `Toml` and other stuff may be better, as that folder seems redundant.

akuleshov7: >>> I prefer this, and I imagine something like TomlEncodingException for encoding
totally agree: `TomlEncodingException`/`TomlDecodingException` sounds very good

akuleshov7: > A further question: If there is only one file in exceptions folder, I think to put it out along with Toml and other stuff may be better, as that folder seems redundant.

Good idea, the directory is redundant. I just automatically used some useless Java practices in Kotlin 😃

Peanuuutz: > Hm, but that will change package name. Instead of ktoml.exceptions.BlaBlaException - you will have ktoml.BlaBlaException. May be better to have it as is...

Then I'll have `TomlEncodingException` in one file, and `TomlDecodingException` another one.

Peanuuutz:
Done.
- Marked `KtomlConf` as `@Deprecated`, and created `TomlConfig` in advance(plan to migrate on 0.4.0).
- Tweak exception classes.
- Changed `field` to `property` because the latter is preferred in Kotlin.

akuleshov7 (approved):
lgtm, let's merge it after #90
and please fix diktat and detekt issues
If you don't mind - I will add you to this repo as a maintainer, so I won't need to approve CI/CD  actions each time

image

Peanuuutz:
diKTat errors this: `public value class ...` should be `value public class ...`, which makes no sense to me. 😶

akuleshov7:
> diKTat errors this: `public value class ...` should be `value public class ...`, which makes no sense to me. 😶

created https://github.com/analysis-dev/diktat/issues/1170 and assigned on @Cheshiriks. I guess it should be fixed in the nearest future. 

@Peanuuutz could you please suppress this case with a @Suppress("WRONG_MULTIPLE_MODIFIERS_ORDER")?

akuleshov7:
If possible - could you please add the section `Acknowledgement` to the readme in this PR and mention something like `Thanks to` NightEule5 Peanuuutz petertrr Olivki and edrd-f  there?

Peanuuutz:
I'm not good at paperwork so feel free to edit. :P

akuleshov7:
> I'm not good at paperwork so feel free to edit. :P

ok, sure 😄

Peanuuutz:
Sure 💪

akuleshov7:
> Sure 💪

oh, it's in fork. Never mind, I will open a new PR :)

Peanuuutz (commented):
Reviewed again. Ready to merge.

akuleshov7:
> Reviewed again. Ready to merge.

Merged, but as you are a maintainer of this repo, can merge by yourself next time. But please use squash-merge

PR #98, opened 2022-02-03
Fix README code formatting

vitusortner:


akuleshov7 (approved):
thank you very much! 
I missed that

PR #117, opened 2022-02-24
Source parser and stream utils

bishiboosh:
The main motivation for this PR is to replicate what's done in https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/json/jvmMain/src/kotlinx/serialization/json/JvmStreams.kt, because I'm using ktoml currently in an Android app and using streams directly without copying them to memory first would be a huge advantage to me.

The modifications done to the lib are the following:
- change `parseStringsToTomlTree` to use a sequence instead of a list.
- add a ktoml-source module, that parses directly from an okio `Source`
- use this module in `ktoml-file`, allowing to avoid copying the file to memory while parsing it
- adding JVM-only methods allowing to parse `InputStream`

I realize that this introduces JVM-only code, so I could spin the stream-related methods into a new project if needs be.


ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/parsers/TomlParser.kt
```diff
@@ -104,18 +132,76 @@ public value class TomlParser(private val config: TomlConfig) {
...
-        if (this.isEmpty()) {
-            return this
-        }
-        // removing all empty lines at the end, to cover empty tables properly
-        while (this.last().isEmptyLine()) {
-            this.removeLast()
-            if (this.isEmpty()) {
-                return this
+    // This code is eavily inspired by the TransformingSequence code in kotlin-lib
+    private fun Sequence<String>.trimEmptyLines(): Sequence<String> {
```
akuleshov7, line 136: Does it look less readable than the example with MutableList? :)

ktoml-source/src/commonMain/kotlin/com/akuleshov7/ktoml/source/TomlSourceReader.kt
```diff
@@ -0,0 +1,96 @@
...
+     *
+     * @param deserializer deserialization strategy
+     * @param source source where toml is stored
+     * @return deserialized object of type T
+     */
+    public fun <T> decodeFromSource(
+        deserializer: DeserializationStrategy<T>,
+        source: Source,
+    ): T {
+        return source.useLines { lines -> decodeFromString(deserializer, lines, config) }
```
akuleshov7, line 34: `it` is fine here instead of `lines`

ktoml-source/src/commonMain/kotlin/com/akuleshov7/ktoml/source/TomlSourceReader.kt
```diff
@@ -0,0 +1,96 @@
...
+     * @param source source where toml is stored
+     * @param tomlTableName fully qualified name of the toml table (it should be the full name -  a.b.c.d)
+     * @return deserialized object of type T
+     */
+    public fun <T> partiallyDecodeFromSource(
+        deserializer: DeserializationStrategy<T>,
+        source: Source,
+        tomlTableName: String,
+    ): T {
+        return source.useLines { lines ->
```
akuleshov7, line 65: same about `it`

ktoml-source/src/commonTest/kotlin/com/akuleshov7/ktoml/file/TomlSourceParserTest.kt
```diff
@@ -0,0 +1,75 @@
...
+            ),
+            Database(
+                "192.168.1.1"
+            )
+        )
+        val buffer = Buffer()
+        buffer.writeUtf8(SIMPLE_EXAMPLE)
+        assertEquals(
+            expected,
+            TomlSourceReader.decodeFromSource(serializer(), buffer)
```
akuleshov7, line 53: Do we still have `decodeFromFile`? 
Is it just an extension for the library or we are removing `decodeFromFile`?

ktoml-source/src/jvmMain/kotlin/com/akuleshov7/ktoml/source/JvmStreams.kt
```diff
@@ -0,0 +1,69 @@
...
+import com.akuleshov7.ktoml.Toml
+import kotlinx.serialization.DeserializationStrategy
+import kotlinx.serialization.serializer
+import okio.source
+import java.io.InputStream
+
+/**
+ * Deserializes TOML from [stream] using UTF-8 encoding to a value of type [T] using [deserializer].
+ */
+public fun <T> Toml.decodeFromStream(
```
akuleshov7, line 12: Do we really need JVM here? don't we want to have a common module for it?

ktoml-source/src/jvmTest/kotlin/com/akuleshov7/ktoml/source/StreamTests.kt
```diff
@@ -0,0 +1,219 @@
...
+                    suiteName = "// DocsCheck",
+                    excludedTests = null,
+                    includedTests = null,
+                    ignoreSaveComments = null
+                ),
+                list2 = listOf(1, 3, 5),
+                warn = WarnConfig(list = listOf("12a", "12f")),
+                list3 = listOf("mystr", "2", "3")
+            ),
+            Toml().decodeFromStream(getTestDataStream("partial_parser_regression.toml"))
```
akuleshov7, line 176: Better to have same syntax as we have for a straight forward decodeFromString:
`Toml.decodeFromString<Bool>("a = \"test\"")`

With a default object companion 

`@ThreadLocal
    public companion object Default : Toml(TomlConfig())`

akuleshov7 (approved):
Generally, looks good and should not affect a lot of people. But API changed, so we will need to take it at least into the `minor` release.

ktoml-source/src/commonMain/kotlin/com/akuleshov7/ktoml/source/SourceUtils.kt
```diff
@@ -0,0 +1,14 @@
+package com.akuleshov7.ktoml.source
+
+import okio.Source
+import okio.buffer
+import okio.use
+
+/**
+ * Read from source one line at a time and passes the lines to the [decoder] function.
+ */
+public inline fun <T> Source.useLines(decoder: (Sequence<String>) -> T): T {
```
akuleshov7, line 10: is it possible to rename it to something more clear? :)
`readAndDecodeLines` or something like this?

akuleshov7:
May be you can also add some small example to the readme? You have added new logic to the API, that can be useful for users. 

I fixed diktat/detekt code style issues and fixed some small typos

bishiboosh:
@akuleshov7 I'll try to look at your comments today, but I may have to delay until next week, I'm a little bit swamped right now

akuleshov7:
@bishiboosh 

You have been waiting for a review for a half of a year (sorry), so I do not think it is a problem for me to wait several weeks 😅

akuleshov7:
Only today I have seen that this PR was opened in the same day when this awful war started one year ago.
A year has passed already and still it has not ended, oh :(

I have opened #188 to finalise this PR: we need also to think how to reduce the number of affected users of the library.

akuleshov7:
And finally, after more than a year I have merged conflicts 🥳

buildSrc/src/main/kotlin/com/akuleshov7/Versions.kt
```diff
@@ -4,8 +4,8 @@
     "PACKAGE_NAME_INCORRECT_PATH")
 
 object Versions {
-    const val KOTLIN = "1.6.21"
+    const val KOTLIN = "1.8.0"
```
nulls: 1.8.20?

akuleshov7: good note, thanks, updated to 1.8.21

PR #119, opened 2022-02-28
Add date-time handling

### LINKED ISSUE #87: Support decoding of date-time format
Should use: https://github.com/Kotlin/kotlinx-datetime for it.

bishiboosh:
Fixes #87


ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/parsers/ValueParserTest.kt
```diff
@@ -18,6 +18,17 @@ class ValueParserTest {
         testTomlValue(Pair("a", "\'false\'"), NodeType.LITERAL_STRING)
     }
 
+    @Test
+    fun dateTimeParsingTest() {
```
akuleshov7, line 22: Seriously, you are the best!

ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/decoders/GeneralDecoderTest.kt
```diff
@@ -6,9 +6,10 @@ import com.akuleshov7.ktoml.exceptions.InvalidEnumValueException
 import com.akuleshov7.ktoml.exceptions.MissingRequiredPropertyException
```
akuleshov7, line 6: can we move it to a separate test? DateTimeDecoder test

ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/parsers/ValueParserTest.kt
bishiboosh: I have a little time right now (probably won't be the case for a few weeks after) so I thought that would be a good idea to do simple issues :)

ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/decoders/GeneralDecoderTest.kt
bishiboosh: sure, I'll do that

bishiboosh:
I just have to precise that local time doesn't exist in kotlinx-serialization, so this case couldn't be handled.

ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/decoders/GeneralDecoderTest.kt
```diff
@@ -432,4 +433,36 @@ class GeneralDecoderTest {
             , Toml.decodeFromString(test)
         )
     }
+
+    @Serializable
+    data class TimeTable(
```
akuleshov7, line 438: Could you please add the negative scenario? When we were not able to parse the value?
And the date inside the string `a = "1979-05-27T00:32:00-07:00"`

bishiboosh: yup, will do that

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
```diff
@@ -25,6 +34,23 @@ public abstract class TomlAbstractDecoder : AbstractDecoder() {
...
+        return descriptor == instantSerializer.descriptor
+                || descriptor == localDateTimeSerializer.descriptor
+                || descriptor == localDateSerializer.descriptor
+    }
+
+    // Cases for date-time types
+    @Suppress("UNCHECKED_CAST")
+    override fun <T> decodeSerializableValue(deserializer: DeserializationStrategy<T>): T {
+        return when (deserializer.descriptor) {
+            instantSerializer.descriptor -> decodePrimitiveType<Instant>() as T
```
akuleshov7, line 47: Why do we need this workaround here? To cast everything to `T`?

bishiboosh: `decodePrimitiveType<Instant>()` returns an `Instant`, where `decodeSerializableValue` expects a T. We know that `Instant` is correct because of the serializer but we still need to cast (this code is inspired by the doc at https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/formats.md#format-specific-types)

bishiboosh:
LocalTime should be available if https://github.com/Kotlin/kotlinx-datetime/issues/57 is fixed, I may do a new PR once a local time implementation lands in kotlinx-datetime

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
akuleshov7: Yeah, that's what I thought

akuleshov7 (approved):

bishiboosh:
I'll probably do the diktat fix somewhere around tomorrow, I'll let you know

bishiboosh:
And no worries for the help, as I said I had some available time, and you're the one to thank for doing all the hard work on this project, I'm just contributing back because you've provided an useful lib :)

PR #121, opened 2022-03-05
Throw error on invalid file extension

bishiboosh:


ktoml-file/src/commonMain/kotlin/com/akuleshov7/ktoml/file/FileUtils.kt
```diff
@@ -20,6 +20,10 @@ import okio.buffer
 internal fun readAndParseFile(tomlFile: String): List<String> {
     try {
         val tomlPath = tomlFile.toPath()
+        val extension = tomlPath.name.substringAfterLast('.', "")
```
akuleshov7, line 23: I am thinking about making it configurable - may be at least move it under some System Property? :)

akuleshov7 (approved):

ktoml-file/src/commonMain/kotlin/com/akuleshov7/ktoml/file/FileUtils.kt
bishiboosh: what do you mean ? making the extension configurable, or the error ?

akuleshov7: I suppose that an error should be configurable with some System Property

PR #122, opened 2022-03-07
Support unicode symbols in strings

### LINKED ISSUE #72: Support Unicode symbols in the strings
Now unicode symbols are not supported. For 
```
"\u0048\u0065\u006C\u006C\u006F World"
```

you will get:
```
According to TOML documentation unknown escape symbols are not allowed.
```

bishiboosh:
Fixes #72


akuleshov7 (approved):
Thank you so much for it! Looks good

PR #125, opened 2022-03-15
Adding tests that show errors in deserialization

akuleshov7:
### What's done:
- test for invalid custom serialization 
- support of custom serialization for small classes (primitive scenario)
- extra complex tests for custom serializers (ignored)


ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/decoders/CustomSerializerTest.kt
```diff
@@ -0,0 +1,86 @@
...
+                """.trimIndent()
+            )
+        )
+    }
+
+    @Serializable
+    data class Settings(val background: SingleProperty, val foreground: SingleProperty)
+
+    @Test
+    @Ignore
```
akuleshov7, line 75: This test does not work, but I actually would have expected it be working

ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/decoders/SurrogateSerializerTest.kt
```diff
@@ -0,0 +1,48 @@
...
+        init {
+            require(r in 0..255 && g in 0..255 && b in 0..255)
+        }
+    }
+
+    @Serializable(with = ColorSerializer::class)
+    class Color(val rgb: Int)
+
+    @Test
+    @Ignore
```
akuleshov7, line 38: This is the example from https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/serializers.md#composite-serializer-via-surrogate

akuleshov7:
@bishiboosh @NightEule5 this is a very small PR with a workaround for custom serializers (with the override of `deserialise()` method), that some users can add.

I hope you don't mind if I merge it without the review as it is very primitive - just a small hack.

I also have found that some more complex cases are not working, so I have added tests and ignored them

PR #128, opened 2022-03-22
Handle commas in string for arrays.

### LINKED ISSUE #115: Array with commas in string values results in error
If there are commas in string values in an array, the deserialization results in an error.

See this simple example :
```toml
myArray = ["Text", "Text, with, commas"]
```

It seems to be because at https://github.com/akuleshov7/ktoml/blob/bcc35fa565c2db3a294334d9448d9e2ea48684d6/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/TomlValue.kt#L267 the comma is selected whether or not it is in a string, but nothing in the TOML spec seems to forbid this

bishiboosh:
Fixes #115


akuleshov7 (approved):
That is great, thank you. Looks very good.

PR #133, opened 2022-03-30
Supporting the decoding of value (inline) classes

akuleshov7:
### What's done:
- Added corresponding logic
- Added tests

PR #134, opened 2022-04-16
Implement writing trees to text

NightEule5:


akuleshov7:
Wow, nice! Will review this week, may be @bishiboosh also would like to have a look :)

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlMainDecoder.kt
```diff
@@ -173,51 +174,63 @@ public class TomlMainDecoder(
...
+    /**
+     * A hack that comes from a compiler plugin to process Inline (value) classes
+     */
+    override fun decodeInline(inlineDescriptor: SerialDescriptor): Decoder =
+            iterateOverStructure(inlineDescriptor, true)
+
     /**
      * this method does all the iteration logic for processing code structures and collections
      * treat it as an !entry point! and the orchestrator of the decoding
      */
```
akuleshov7, line 186: very strange that we see changes here... it looks like they are coming from https://github.com/akuleshov7/ktoml/pull/133

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/TomlArrayOfTables.kt
```diff
@@ -65,6 +66,56 @@ public class TomlArrayOfTables(
             (0..index).joinToString(".") { sectionsList[it] }
         }
     }
+
+    override fun TomlEmitter.writeHeader(headerKey: TomlKey, config: TomlConfig) {
```
akuleshov7, line 70: I am starting to think about splitting this config to separate configs: TomlInputConfig and TomlOutputConfig, what do you think about that?

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/TomlKey.kt
```diff
@@ -38,4 +41,29 @@ public class TomlKey(public val rawContent: String, public val lineNo: Int) {
         }
         return false
     }
+
+    public fun write(emitter: TomlEmitter, config: TomlConfig) {
+        val keys = keyParts
+
+        if (keys.isEmpty() || keys.any(String::isEmpty)) {
```
akuleshov7, line 48: Actually we store `rawContent` in this class already. 
But your logic is more straight-forward, as you work directly with the internal representation of Node. 

What do you think - what is the best thing here to use? `rawContent` or `keyParts`?

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/TomlArrayOfTables.kt
```diff
@@ -65,6 +66,56 @@ public class TomlArrayOfTables(
...
+                child.write(emitter = this, config, multiline)
+
+                dedent()
+
+                if (i < last) {
+                    emitNewLine()
+
+                    // Primitive pairs have a single newline after, except when a
+                    // table follows.
+                    if (child !is TomlKeyValuePrimitive ||
```
akuleshov7, line 109: could you please put it to a single line, it is not long

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/TomlArrayOfTables.kt
```diff
@@ -65,6 +66,56 @@ public class TomlArrayOfTables(
...
+                            children[i + 1] is TomlTable) {
+                        emitNewLine()
+                    }
+                }
+            } else {
+                child.write(emitter = this, config, multiline)
```
akuleshov7, line 115: do we need to emit a newline here?

ktoml-core/src/linuxX64Main/kotlin/com/akuleshov7/ktoml/utils/RegexesLinux.kt
```diff
@@ -0,0 +1,9 @@
+/**
+ * Regexes used in tree writing.
+ */
+
+package com.akuleshov7.ktoml.utils
+
+internal actual val controlCharacterRegex = commonControlCharacterRegex
```
akuleshov7, line 7: actually there is no need to create 3 directories for this logic. You can create a gradle task: `nativeMain`. Here is a good example of this: https://github.com/analysis-dev/save-cli/blob/c4414a8caa023900b62ef2ab4b8628ec35c8510f/save-cli/build.gradle.kts


```
        val nativeMain by creating {
            dependsOn(commonMain)
            dependencies {
                implementation(projects.saveCore)
                implementation(libs.kotlinx.serialization.properties)
            }
        }
        saveTarget.forEach {
            getByName("${it.name}Main").dependsOn(nativeMain)
        }
```

akuleshov7: we can even have naiveNonJvm that will join JS, linux, mac, win and many new targets

ktoml-core/src/jvmMain/kotlin/com/akuleshov7/ktoml/utils/RegexesJvm.kt
```diff
@@ -0,0 +1,25 @@
+/**
+ * Regexes used in tree writing.
+ */
+
+package com.akuleshov7.ktoml.utils
+
+import kotlin.text.RegexOption.COMMENTS
+
+internal actual val controlCharacterRegex = Regex("""[\p{Cc}\x7F&&[^\t]]""")
```
akuleshov7, line 9: could you please add comment of what kind of regex you have here and why

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/TomlInlineTable.kt
```diff
@@ -75,4 +76,31 @@ public class TomlInlineTable(
...
+            if (i > 0) {
+                emitter.emitElementDelimiter()
+            }
+
+            emitter.emitWhitespace()
+
+            pair.write(emitter, config)
+        }
+
+        emitter.emitWhitespace()
```
akuleshov7, line 103: I have a small suggestion here - may be we will change emmiter API in the following way:
- all methods will return `this` in a functional way
- so we would be able to use it in the functional way:
```
        emitter.emitWhitespace()
                    .endInlineTable()
                    .enitNewLine()
... 
```

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/TomlStubEmptyNode.kt
```diff
@@ -14,4 +15,10 @@ public class TomlStubEmptyNode(lineNo: Int, config: TomlConfig = TomlConfig()) :
     config
 ) {
     override val name: String = EMPTY_TECHNICAL_NODE
+
+    override fun write(
+        emitter: TomlEmitter,
+        config: TomlConfig,
+        multiline: Boolean
+    ) { }
```
akuleshov7, line 23: strange that diktat is not raising a warning here, could you please add a comment inside this block, something like: `no need to write anything for a stub node`?

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/TomlTable.kt
```diff
@@ -26,6 +28,69 @@ public abstract class TomlTable(
...
+
+    override fun write(
+        emitter: TomlEmitter,
+        config: TomlConfig,
+        multiline: Boolean
+    ) {
+        // Todo: Option to explicitly define super tables?
+        // Todo: Support dotted key-value pairs (i.e. a.b.c.d = 7)
+
+        val children = children
```
akuleshov7, line 40: this caching looks a little bit strange - may be use children from `this`?

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/TomlTable.kt
```diff
@@ -26,6 +28,69 @@ public abstract class TomlTable(
...
+
+        val key = TomlKey(fullTableName, 0)
+
+        val firstChild = children.first()
+
+        if (isExplicit(firstChild) && type == TableType.PRIMITIVE) {
+            emitter.writeHeader(key, config)
+
+            if (firstChild !is TomlStubEmptyNode) {
+                emitter.emitNewLine()
```
akuleshov7, line 50: If first child is a stub, this means, that it is a table without children: 
```
[a]
[b]
[c]
```

I suppose we can have a newline after it

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/utils/Regexes.kt
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+package com.akuleshov7.ktoml.utils
+
+internal expect val controlCharacterRegex: Regex
+
+internal expect val unescapedBackslashRegex: Regex
+
+internal val commonControlCharacterRegex =
+        Regex("""[\x00-\x08\x0A-\x1F\x7F\x80-\x9F]""")
```
akuleshov7, line 12: please add comments about these regexes

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/writers/TomlEmitter.kt
```diff
@@ -207,6 +207,32 @@ public abstract class TomlEmitter(config: TomlConfig) {
...
+     * @param instant
+     */
+    public fun emitValue(instant: Instant): Unit = emit(instant.toString())
+
+    /**
+     * Emits a [LocalDateTime] value.
+     *
+     * @param dateTime
+     */
+    public fun emitValue(dateTime: LocalDateTime): Unit = emit(dateTime.toString())
```
akuleshov7, line 222: all these methods do the same - convert value `toString`, may be unify it with generic method?

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/TomlValue.kt
```diff
@@ -227,6 +352,19 @@ internal constructor(
...
+    override fun write(
+        emitter: TomlEmitter,
+        config: TomlConfig,
+        multiline: Boolean
+    ) {
+        when (val content = content) {
+            is Instant -> emitter.emitValue(content)
+            is LocalDateTime -> emitter.emitValue(content)
+            is LocalDate -> emitter.emitValue(content)
+            else -> { }
```
akuleshov7, line 364: may be throw an exception here?

akuleshov7 (approved):

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlMainDecoder.kt
NightEule5: I rebased when new changes were made to the main branch, which seems to be the wrong thing to do. Should I have merged the main branch in this one?

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/TomlArrayOfTables.kt
NightEule5: Sure, that'd be useful

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/TomlKey.kt
NightEule5: I see what you're saying, `rawContent.isEmpty()` would work. I think `keyParts` makes more sense though. As you said during encoding the content won't be available unless we format it.

Maybe it'd be better to pass `keyParts` in an internal primary constructor? Something like this perhaps:
```kotlin
public class TomlKey internal constructor(
    internal val keyParts: List<String>,
   public val lineNo: Int
) {
    constructor(rawContent: String, lineNo: Int) : this(rawContent.splitKeyToTokens(lineNo), lineNo)
}
```

For now I was basically working with the existing API with as minimal changes as possible

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/TomlArrayOfTables.kt
NightEule5: Otherwise we have no blank line between blocks, so the output would look like this:

```toml
[[fruits]]
    a = "apple"
    b = "qwerty"
[[fruits]]
    a = "banana"
    b = "qwerty"
[[fruits]]
    a = "plantain"
    b = "qwerty"
```

Or is that the desired output?

ktoml-core/src/jvmMain/kotlin/com/akuleshov7/ktoml/utils/RegexesJvm.kt
NightEule5: `&&` is a class intersection and `\p{Cc}` is a unicode category, both only present on the JVM. The less compact version for all other platforms is clearer though, should we use that instead and avoid multiple variations? I'll document it for now

NightEule5: Actually, I'll just remove it. Adds unnecessary complexity to the project

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/TomlTable.kt
NightEule5: Yep, I have a habit of doing this to avoid accessing a field more than once, but I agree it looks weird. I'll change it

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/writers/TomlEmitter.kt
NightEule5: Well, a public generic method would allow passing potentially invalid values (the emitter API is public after all). The only common class between these is `Comparable`. Does that matter?

A private method like

```kotlin
private fun emitValue(value: Any) = emit(value.toString())
```

is doesn't have these issues though.

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/TomlTable.kt
NightEule5: Yep, that was intentional. We could maybe have some config values to effect formatting stuff like this, utilizing the separate input and output config classes you suggested

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/TomlValue.kt
```diff
@@ -175,6 +257,31 @@ internal constructor(
...
+                    '\n' -> "\\n"
+                    '\u000C' -> "\\f"
+                    '\r' -> "\\r"
+                    else -> {
+                        val code = char.code
+
+                        val hexDigits = code.toString(HEX_RADIX)
+
+                        "\\$SIMPLE_UNICODE_PREFIX${
+                            hexDigits.padStart(SIMPLE_UNICODE_LENGTH, '0')
```
bishiboosh, line 274: I think if you only use the `\u` prefix, then it's going to be problematic for complex unicode symbols. For example, 🔥 is `U+1F525`, and would be written as `\U0001F525` in TOML. A simple method would just be to check `hexDigits` length and see which prefix to use

bishiboosh (changes requested):

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/TomlKey.kt
akuleshov7: yeah, let's leave that as is and will change while will be making encoding

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/TomlValue.kt
NightEule5: This code is really only supposed to touch control characters though, which are well within the `\u` range

bishiboosh: oh alright, my bad, didn't understand that part properly

bishiboosh (approved):

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/writers/TomlEmitter.kt
akuleshov7: yeah, ok, let it be :)

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlMainDecoder.kt
akuleshov7: Right now everything looks fine actually. I simply merged changes from ktoml main branch to your branch.

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/TomlArrayOfTables.kt
akuleshov7: I will create an issue on that https://github.com/akuleshov7/ktoml/issues/135

NightEule5:
Ok, are we good to merge then?

Issue #114, opened 2022-02-22
Annotation: annotations that are needed for the encoding 


akuleshov7:
- EncodeDefault 
- LiteralString
- AddComment

NightEule5:
Btw, apparently `EncodeDefault` already exists in kotlinx.serialization

akuleshov7:
> Btw, apparently `EncodeDefault` already exists in kotlinx.serialization
yeah, I missed that, thank you

PR #139, opened 2022-05-05
Add serialization annotations

NightEule5:
I've added the annotations suggested in #114, as well as a few more I thought would be desired: `TomlInlineTable`, `TomlInteger`, and `TomlMultiline`


bishiboosh (approved):

akuleshov7 (approved):
That's awesome!

PR #140, opened 2022-05-05
Store comments in the tree

### LINKED ISSUE #136: Store comments in the tree
For encoding, we should support adding comments to the file. Since an `AddComment` annotation is suggested in #114, I assume this is an intended feature. We'll need to add comments to the tree

Comments are likely to be "attached" to other elements, like:

```toml
# Comment
key = "value" # Other comment
```

So I think it makes sense to apply comments to a `TomlNode` instance, rather than being nodes themselves.

Something like this:

```kotlin
public sealed class TomlNode(
    public open val content: String,
    public open val lineNo: Int,
    public val comments: List<String>,
    public val inlineComment: String,
    public open val config: TomlConfig = TomlConfig()
)
```

Or:

```kotlin
data class TomlComments(
    public val comments: List<String>,
    public val inlineComments: String
)

public sealed class TomlNode(
    public open val content: String,
    public open val lineNo: Int,
    public val comments: TomlComments,
    public open val config: TomlConfig = TomlConfig()
)
```

We could also optionally retain or ignore comments during parsing with an `ignoreComments` property in the config

NightEule5:
Fixes #136

### DIFF
--- a/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/TomlConfig.kt
+++ b/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/TomlConfig.kt
@@ -23,8 +23,6 @@ public class KtomlConf(
  * @property allowEscapedQuotesInLiteralStrings - a control to allow/prohibit escaping of single quotes in literal strings
  * @property indentation - the number of spaces in the indents for the serialization
  * @property allowEmptyToml - controls if empty toml can be processed, if false - will throw an exception
- * @property ignoreComments - controls whether comments are ignored on deserialization. If true, comments
- * will be present in the tree.
  */
 public open class TomlConfig(
     public val ignoreUnknownNames: Boolean = false,
@@ -33,7 +31,6 @@ public open class TomlConfig(
     public val allowEscapedQuotesInLiteralStrings: Boolean = true,
     public val indentation: Indentation = Indentation.FOUR_SPACES,
     public val allowEmptyToml: Boolean = true,
-    public val ignoreComments: Boolean = true,
 ) {
     /**
      * @property value - string with indents, used for the formatting of serialization
--- a/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/TomlConfig.kt
+++ b/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/TomlConfig.kt
@@ -23,6 +23,8 @@ public class KtomlConf(
  * @property allowEscapedQuotesInLiteralStrings - a control to allow/prohibit escaping of single quotes in literal strings
  * @property indentation - the number of spaces in the indents for the serialization
  * @property allowEmptyToml - controls if empty toml can be processed, if false - will throw an exception
+ * @property ignoreComments - controls whether comments are ignored on deserialization. If true, comments
+ * will be present in the tree.
  */
 public open class TomlConfig(
     public val ignoreUnknownNames: Boolean = false,
@@ -31,6 +33,7 @@ public open class TomlConfig(
     public val allowEscapedQuotesInLiteralStrings: Boolean = true,
     public val indentation: Indentation = Indentation.FOUR_SPACES,
     public val allowEmptyToml: Boolean = true,
+    public val ignoreComments: Boolean = true,
 ) {
     /**
      * @property value - string with indents, used for the formatting of serialization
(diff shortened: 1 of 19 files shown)


ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/TomlConfig.kt
```diff
@@ -31,6 +33,7 @@ public open class TomlConfig(
     public val allowEscapedQuotesInLiteralStrings: Boolean = true,
     public val indentation: Indentation = Indentation.FOUR_SPACES,
     public val allowEmptyToml: Boolean = true,
+    public val ignoreComments: Boolean = true,
```
akuleshov7, line 36: do we really need to ignore comments from the input?

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/parsers/TomlParser.kt
```diff
@@ -40,18 +40,34 @@ public value class TomlParser(private val config: TomlConfig) {
         // here we always store the bucket of the latest created array of tables
         var latestCreatedBucket: TomlArrayOfTablesElement? = null
 
+        val comments: MutableList<String> = mutableListOf()
```
akuleshov7, line 43: We should be very careful here. Do you link your comments from previous lines to the next token?

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/parsers/TomlParser.kt
```diff
@@ -40,18 +40,34 @@ public value class TomlParser(private val config: TomlConfig) {
...
+            if (line.isComment()) {
+                if (!config.ignoreComments) {
+                    comments += line.trimCommentLine()
+                }
+            } else if (!line.isEmptyLine()) {
+                // Parse the inline comment if any
+                val inlineComment = if (config.ignoreComments) {
+                    ""
+                } else {
+                    when (val hashIndex = line.indexOf('#')) {
```
akuleshov7, line 58: Let's move it to some method? Parsing function is very long already

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/parsers/TomlParser.kt
```diff
@@ -40,18 +40,34 @@ public value class TomlParser(private val config: TomlConfig) {
...
+                }
+
                 if (line.isTableNode()) {
                     if (line.isArrayOfTables()) {
                         // TomlArrayOfTables contains all information about the ArrayOfTables ([[array of tables]])
                         val tableArray = TomlArrayOfTables(line, lineNo, config)
                         val arrayOfTables = tomlFileHead.insertTableToTree(tableArray, latestCreatedBucket)
                         // creating a new empty element that will be used as an element in array and the parent for next key-value records
-                        val newArrayElement = TomlArrayOfTablesElement(lineNo, config)
+                        val newArrayElement = TomlArrayOfTablesElement(lineNo, comments.toList(), inlineComment, config)
```
akuleshov7, line 70: May be accept Mutable list in the Toml nodes? Instead of making `toList` everywhere?
Or not... As you wish 😄

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/parsers/TomlParser.kt
```diff
@@ -118,6 +136,9 @@ public value class TomlParser(private val config: TomlConfig) {
         return this
     }
 
+    private fun String.trimCommentLine() =
```
akuleshov7, line 139: Actually we already have some logic related to trimming of comments. 
But yes, it looks much better here. May be you can remove all usages of: `content.findBeginningOfTheComment` though?

akuleshov7 (approved):
Approved, but pls see some small remarks

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/parsers/TomlParser.kt
NightEule5: Yes, exactly

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/parsers/TomlParser.kt
NightEule5: Well, `toList` is there to copy the elements to a readonly list. If we passed the `comments` list itself the comments would be modified when the next element is being parsed

Maybe it's better to create a new list every element instead

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/parsers/TomlParser.kt
NightEule5: I'll replace it with a `takeBeforeComment` function, since it's always used to trim the comment away

akuleshov7: awesome, totally agree

NightEule5:
@akuleshov7 I went ahead and merged since the changes were approved, hope you don't mind

Issue #145, opened 2022-06-04
Add extra targets and releases for native platforms


akuleshov7:
At least: 
- macOSArm64 (M1)
- Android/iOS/WatchOS ?

Wavesonics:
+1 for mac support

Android should be covered by the JVM target.

Does the existing Arm target cover ios?

PR #148, opened 2022-06-16
Add Ios target

Wavesonics:
Adds an iOS meta target,

I'm not sure if there is more to be done on the publishing side? But this compiles for me.

### DIFF
--- a/buildSrc/src/main/kotlin/com/akuleshov7/Versions.kt
+++ b/buildSrc/src/main/kotlin/com/akuleshov7/Versions.kt
@@ -6,6 +6,6 @@
 object Versions {
     const val KOTLIN = "1.6.21"
     const val JUNIT = "5.7.1"
-    const val OKIO = "3.0.0"
+    const val OKIO = "3.1.0"
     const val SERIALIZATION = "1.3.2"
 }
--- a/ktoml-core/src/iosMain/kotlin/com/akuleshov7/ktoml/utils/UtilsIos.kt
+++ b/ktoml-core/src/iosMain/kotlin/com/akuleshov7/ktoml/utils/UtilsIos.kt
@@ -1,3 +1,7 @@
+/**
+ * Specific implementation for utilities
+ */
+
 package com.akuleshov7.ktoml.utils
 
 @Suppress("MAGIC_NUMBER")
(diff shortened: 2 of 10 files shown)


buildSrc/src/main/kotlin/com/akuleshov7/Versions.kt
```diff
@@ -6,6 +6,6 @@
 object Versions {
     const val KOTLIN = "1.6.21"
     const val JUNIT = "5.7.1"
-    const val OKIO = "3.0.0"
+    const val OKIO = "3.1.0"
```
Wavesonics, line 9: This added the iOS common target

ktoml-core/src/iosMain/kotlin/com/akuleshov7/ktoml/utils/UtilsIos.kt
```diff
@@ -0,0 +1,11 @@
...
+
+@Suppress("MAGIC_NUMBER")
+internal actual fun StringBuilder.appendCodePointCompat(codePoint: Int): StringBuilder = when (codePoint) {
+    in 0 until Char.MIN_SUPPLEMENTARY_CODE_POINT -> append(codePoint.toChar())
+    in Char.MIN_SUPPLEMENTARY_CODE_POINT..Char.MAX_CODE_POINT -> {
+        append(Char.MIN_HIGH_SURROGATE + ((codePoint - 0x10000) shr 10))
+        append(Char.MIN_LOW_SURROGATE + (codePoint and 0x3ff))
+    }
+    else -> throw IllegalArgumentException()
+}
```
Wavesonics, line 11: I just copied this code from the linux version, is there anything platform specific needed?

akuleshov7: Nope, I do not even think that this method is really a cross-platform code and cannot be in common :)

akuleshov7:
I will make a release after https://github.com/akuleshov7/ktoml/issues/145 I will add all other platforms this week

akuleshov7:
@Wavesonics thank you for your contribution, merged

PR #153, opened 2022-10-20
Implement encoding

### LINKED ISSUE #11: Support encoding of toml format
Decoder has the first priority as the Toml usually is used for configuration files, but encoders should also be supported

### LINKED ISSUE #151: Implement `encodeToString`
This would be really useful in general

NightEule5:
Fixes #151
Fixes #11

There are a few problems with this still:
- The `TomlComments` annotation throws exceptions related to its `inline` default parameter. Nothing we can do about this, it's an issue with kotlinx.serialization that hasn't been fixed yet afaik
- The `isSynthetic` property in `TomlTable` was made mutable, but that's a breaking change. I've been rewriting the encoders to defer table creation to the `endStructure` function to avoid this
- Some formatting quirks persist with regards to indentation
- Annotations are inaccessible from property types, i.e. `val string: @TomlLiteral String` produces a normal string. Looks like another limitation of kotlinx.serialization, but I'm not sure how useful that would be anyway

### DIFF
--- a/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/Toml.kt
+++ b/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/Toml.kt
@@ -7,14 +7,13 @@ import com.akuleshov7.ktoml.parsers.TomlParser
 import com.akuleshov7.ktoml.tree.TomlFile
 import com.akuleshov7.ktoml.utils.findPrimitiveTableInAstByName
 import com.akuleshov7.ktoml.writers.TomlWriter
-
-import kotlin.native.concurrent.ThreadLocal
 import kotlinx.serialization.DeserializationStrategy
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.SerializationStrategy
 import kotlinx.serialization.StringFormat
 import kotlinx.serialization.modules.EmptySerializersModule
 import kotlinx.serialization.modules.SerializersModule
+import kotlin.native.concurrent.ThreadLocal
 
 /**
  * Toml class - is a general entry point in the core,
@@ -61,7 +60,7 @@ public open class Toml(
     }
 
     override fun <T> encodeToString(serializer: SerializationStrategy<T>, value: T): String {
-        val toml = TomlMainEncoder.encode(serializer, value, inputConfig, outputConfig)
+        val toml = TomlMainEncoder.encode(serializer, value, inputConfig, outputConfig, serializersModule)
 
         return tomlWriter.writeToString(file = toml)
     }
--- a/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/Toml.kt
+++ b/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/Toml.kt
@@ -7,13 +7,13 @@ import com.akuleshov7.ktoml.parsers.TomlParser
 import com.akuleshov7.ktoml.tree.TomlFile
 import com.akuleshov7.ktoml.utils.findPrimitiveTableInAstByName
 import com.akuleshov7.ktoml.writers.TomlWriter
+import kotlin.native.concurrent.ThreadLocal
 import kotlinx.serialization.DeserializationStrategy
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.SerializationStrategy
 import kotlinx.serialization.StringFormat
 import kotlinx.serialization.modules.EmptySerializersModule
 import kotlinx.serialization.modules.SerializersModule
-import kotlin.native.concurrent.ThreadLocal
 
 /**
  * Toml class - is a general entry point in the core,
--- a/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/Toml.kt
+++ b/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/Toml.kt
@@ -27,7 +27,7 @@ import kotlinx.serialization.modules.SerializersModule
 public open class Toml(
     protected val inputConfig: TomlInputConfig = TomlInputConfig(),
     protected val outputConfig: TomlOutputConfig = TomlOutputConfig(),
-    override val serializersModule: SerializersModule = EmptySerializersModule
+    override val serializersModule: SerializersModule = EmptySerializersModule(),
 ) : StringFormat {
     // parser and writer are created once after the creation of the class, to reduce
     // the number of created parsers and writers for each toml
@@ -39,7 +39,7 @@ public open class Toml(
     )
     public constructor(
         config: TomlConfig,
-        serializersModule: SerializersModule = EmptySerializersModule
+        serializersModule: SerializersModule = EmptySerializersModule(),
     ) : this(
         config.input,
         config.output,
--- a/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/Toml.kt
+++ b/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/Toml.kt
@@ -1,6 +1,7 @@
 package com.akuleshov7.ktoml
 
 import com.akuleshov7.ktoml.decoders.TomlMainDecoder
+import com.akuleshov7.ktoml.encoders.TomlMainEncoder
 import com.akuleshov7.ktoml.exceptions.MissingRequiredPropertyException
 import com.akuleshov7.ktoml.parsers.TomlParser
 import com.akuleshov7.ktoml.tree.TomlFile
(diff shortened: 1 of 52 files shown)


akuleshov7:
Awesome! I just yesterday starting doing it 😂

NightEule5:
haha

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/Toml.kt
```diff
@@ -60,7 +60,9 @@ public open class Toml(
     }
 
     override fun <T> encodeToString(serializer: SerializationStrategy<T>, value: T): String {
-        TODO("Not yet implemented")
+        val toml = TomlMainEncoder.encode(serializer, value, inputConfig, outputConfig, serializersModule)
```
akuleshov7, line 63: why do we need inputConfig right here? I thought we will split it

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/encoders/TomlAbstractEncoder.kt
```diff
@@ -0,0 +1,388 @@
...
+        }
+    }
+
+    override fun encodeEnum(enumDescriptor: SerialDescriptor, index: Int) {
+        encodeString(enumDescriptor.getElementName(index))
+    }
+
+    override fun encodeLong(value: Long) {
+        if (attributes.intRepresentation != DECIMAL) {
+            throw UnsupportedEncodingFeatureException(
```
akuleshov7, line 104: in abstractDecoder we have a method `invalidType` may be make it common and use it here too?

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/encoders/TomlAbstractEncoder.kt
```diff
@@ -0,0 +1,388 @@
...
+        }
+    }
+
+    override fun encodeByte(value: Byte): Nothing = invalidType("Byte", "Long", value)
+    override fun encodeShort(value: Short): Nothing = invalidType("Short", "Long", value)
+    override fun encodeInt(value: Int): Nothing = invalidType("Int", "Long", value)
+    override fun encodeFloat(value: Float): Nothing = invalidType("Float", "Double", value)
+    override fun encodeChar(value: Char): Nothing = invalidType("Char", "String", value)
+
+    // Todo: Do we really want to make these invalid?
```
akuleshov7, line 165: yeah, I also started thinking that following toml specification in a so straight-forward way can be a problem

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/encoders/TomlAbstractEncoder.kt
```diff
@@ -0,0 +1,388 @@
...
+    private fun invalidType(
```
akuleshov7, line 166: better to make this method common for encoder and decoder...

akuleshov7:
will be great also to have some examples in readme, but I can do it

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/encoders/TomlAbstractEncoder.kt
```diff
@@ -0,0 +1,388 @@
...
+     * are represented.
+     * @property isInline Marks subsequent table-like elements as inline. Tables
+     * will be written as inline tables.
+     * @property comments Comment lines to be prepended before the next element.
+     * @property inlineComment A comment to be appended to the end of the next
+     * element's line.
+     * @property isImplicit Whether the current property is implicitly defined in
+     * its child, i.e. the table `[a]` in `[a.b]`.
+     */
+    public data class Attributes(
```
akuleshov7, line 336: why do we need this class? also - can we move it to a separate file, I see that it used in several places

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/TomlValue.kt
```diff
@@ -132,10 +133,6 @@ internal constructor(
                             "Control characters (excluding tab) are not permitted" +
                                     " in literal strings."
                         )
-                    '\\' in this ->
```
akuleshov7, line 135: was this intentional?

akuleshov7 (approved):
actually implementation looks very good for a POC, we can polish it later, 
but need to create tickets for enhancements

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/Toml.kt
NightEule5: We did, but nodes still require it in the constructor

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/TomlValue.kt
NightEule5: Yep, not sure why it was there to begin with tbh. Literal strings are supposed to interpret "\\" as a normal character rather than an escape, except our non-spec single quote escapes which are handled already

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/encoders/TomlAbstractEncoder.kt
NightEule5: It just makes keeping track of and passing element information easier. Sure, I'll move it to a separate file

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/Toml.kt
akuleshov7: Ah, I got it. We create tree first even for encoding, sure

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/encoders/TomlAbstractEncoder.kt
akuleshov7: May be later? Can you please create a separate issue? Let’s not take too much in one PR :)

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/encoders/TomlAbstractEncoder.kt
akuleshov7: Got it, ok, let’s move to a separate class, right now this one looks overloaded

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/TomlValue.kt
akuleshov7: Could you please add your example to tests? Looks like I missed that

akuleshov7:
@NightEule5 btw have you seen this proeblem with polymorphism ? 


44C185B5-AE8F-43C5-A795-B906FC6B175A

NightEule5:
> @NightEule5 btw have you seen this proeblem with polymorphism ?
> 
> If needed - we can skip polymorphism implementation by for now

I'd only been running JVM tests locally, so I didn't encounter it. Apparently on Native you have to explicitly mark enums as `@Serializable`, or else you get this error:

> kotlinx.serialization.SerializationException: Serializer for class 'Enum' is not found.
Mark the class as @Serializable or provide the serializer explicitly.
On Kotlin/Native explicitly declared serializer should be used for interfaces and enums without @Serializable annotation
kotlinx.serialization.SerializationException: Serializer for class 'Enum' is not found.

Easy fix

NightEule5:
The last test fail is on JS, in cases where a double has no fractional component. So the doc test was failing here:

```toml
# Should be this
temp_targets = { cpu = 79.5, case = 72.0 }

# Actual output was an integer
temp_targets = { cpu = 79.5, case = 72 }
```

akuleshov7:
> The last test fail is on JS, in cases where a double has no fractional component. So the doc test was failing here:
> 
> ```toml
> # Should be this
> temp_targets = { cpu = 79.5, case = 72.0 }
> 
> # Actual output was an integer
> temp_targets = { cpu = 79.5, case = 72 }
> ```

Should be an easy fix, but we can do it later if you wish

NightEule5:
>  Should be an easy fix, but we can do it later if you wish in a separate PR

Sure

ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/encoders/TomlDocsEncoderTest.kt
```diff
@@ -0,0 +1,162 @@
...
+                    .toInstant(TimeZone.UTC)
+    )
+
+    @Serializable
+    data class Database(
+        val enabled: Boolean = true,
+        val ports: List<Long> = listOf(8000, 8001, 8002),
+        @Serializable(with = DataSerializer::class)
+        val data: List<List<@Contextual Any>> = listOf(listOf("delta", "phi"), listOf(3.14)),
+        // FixMe: following code is not working on js target:
```
akuleshov7, line 44: Issue with JS port: invalid serialization of double

PR #158, opened 2022-10-31
Fix Double encoding on JS

### LINKED ISSUE #155: Issue with JS port: invalid serialization of double
Issue with JS port: invalid serialization of double

_Originally posted by @akuleshov7 in https://github.com/akuleshov7/ktoml/pull/153#discussion_r1008911491_

NightEule5:
Fixes #155

### DIFF
--- a/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/writers/TomlEmitter.kt
+++ b/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/writers/TomlEmitter.kt
@@ -223,7 +223,12 @@ public abstract class TomlEmitter(config: TomlOutputConfig) {
                 float.isNaN() -> "nan"
                 float.isInfinite() ->
                     if (float > 0) "inf" else "-inf"
-                else -> float.toString()
+                else -> {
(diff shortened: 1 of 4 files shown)


ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/writers/TomlEmitter.kt
```diff
@@ -223,7 +223,12 @@ public abstract class TomlEmitter(config: TomlOutputConfig) {
                 float.isNaN() -> "nan"
                 float.isInfinite() ->
                     if (float > 0) "inf" else "-inf"
-                else -> float.toString()
+                else -> {
```
akuleshov7, line 226: May be need to report/fix in Kotlin? What do you think? Or not a bug

akuleshov7 (approved):

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/writers/TomlEmitter.kt
NightEule5: I think it is. I'm guessing something to do with JS's dynamic typing

Issue #163, opened 2022-11-18
Support encoding/decoding non-Long numbers


aSemy:
Currently when I try to decode an integer to a `Int`, I get an error. This is inconvenient, and confusing.

```
Exception in thread "main" com.akuleshov7.ktoml.exceptions.IllegalTypeException: 
Line 2: <Int> type is not allowed by toml specification, use <Long> instead (key = port; value = 8080)
```

```kotlin
import com.akuleshov7.ktoml.Toml
import kotlinx.serialization.Serializable

fun main() {
  Toml.decodeFromString(ServerConfig.serializer(), """
    host = "localhost"
    port = 8080
  """.trimIndent())
}

@Serializable
data class ServerConfig(
  val host: String,
  val port: Int,
)
```

### Comparison with Kotlinx Serialization

This is confusing because it does not match the behaviour of Kotlinx Serialization JSON. Technically, the JSON spec states that any number can be of any magnitude or precision (so basically KxS should only support `BigDecimal` or `BigInteger`), but KxS allows for numbers to be en/decoded with any numeric type - `Long`, `Short`, `Double`, `Float` - including unsigned numbers.

```kotlin
import kotlinx.serialization.Serializable
import kotlinx.serialization.decodeFromString
import kotlinx.serialization.json.Json

@Serializable
data class ServerConfig(
  val host: String,
  val port: Int, // Int is allowed
)

fun main() {
  val config: ServerConfig = Json.decodeFromString(
    """
    {
      "host": "localhost",
      "port": 123213131321321312313213213132123132131
    }
  """.trimIndent()
  )
}
```

KxS throws a runtime error when it encounters an invalid number 

```
Exception in thread "main" kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 35: Numeric value overflow at path: $.port
JSON input: {
  "host": "localhost",
  "port": 123213131321321312313213213132123132131
}
```

#### Custom Serializer 
I have tried to work around this with a custom serializer, but I get the same error.

```kotlin
internal typealias IntAsLong = @Serializable(with = IntAsLongSerializer::class) Int

private object IntAsLongSerializer : KSerializer<Int> {
  override val descriptor: SerialDescriptor =
    PrimitiveSerialDescriptor("IntAsLongSerializer", PrimitiveKind.LONG)

  override fun serialize(encoder: Encoder, value: Int) = encoder.encodeLong(value.toLong())

[...]

PR #164, opened 2022-11-21
quick POC at non-Long numerics (encoding and decoding)

aSemy:
A quick stab at showing how non-Long numbers could be encoded and decoded. 

#163 

This is a very quick draft, I haven't spent a lot of time on it. I'm just sharing because I think the general approach is good.

This approach is based off the Kotlinx Serialization approach to handling JSON numbers

https://github.com/Kotlin/kotlinx.serialization/blob/dc950f5098162a3f8dd742d04b95f9a0405470e1/formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt#L281-L290

First the numbers are parsed as a Long or Double, as is already the case. Then ktoml will try to manually convert these numbers to the required type, but first verifying that the numbers are in range.

If you wanted to verify based on the raw string content, you could perhaps implement a function based on how KxS parses numbers: https://github.com/Kotlin/kotlinx.serialization/blob/dc950f5098162a3f8dd742d04b95f9a0405470e1/formats/json/commonMain/src/kotlinx/serialization/json/internal/lexer/AbstractJsonLexer.kt#L533-L584

### DIFF
--- a/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
+++ b/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
@@ -2,6 +2,8 @@ package com.akuleshov7.ktoml.decoders
 
 import com.akuleshov7.ktoml.exceptions.CastException
 import com.akuleshov7.ktoml.exceptions.IllegalTypeException
+import com.akuleshov7.ktoml.exceptions.ParseException
+import com.akuleshov7.ktoml.exceptions.TomlDecodingException
 import com.akuleshov7.ktoml.tree.nodes.TomlKeyValue
 import kotlinx.datetime.Instant
 import kotlinx.datetime.LocalDate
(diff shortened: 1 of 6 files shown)


ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
```diff
@@ -20,12 +19,46 @@ public abstract class TomlAbstractDecoder : AbstractDecoder() {
...
+    override fun decodeInt(): Int = decodePrimitiveIntegerType { it.toInt() }
+
+    override fun decodeFloat(): Float {
+        val keyValue = decodeKeyValue()
+
+        val doubleValue = keyValue.castOrThrow<Double>()
+
+        val floatValue = doubleValue.toFloat()
+
+        if (floatValue.isInfinite()) {
```
NightEule5, line 35: Json apparently doesn't have any way to express infinity and NaN, so I guess that's why KxS has this exception. Toml does have `inf` and `nan` though

akuleshov7:
@aSemy I will merge your part into #171 , I have made a simple part, could you please review?

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
akuleshov7: Looks like I actually missed that, need to support it...

aSemy:
> @aSemy I will merge your part into #171 , I have made a simple part, could you please review?

Of course! Thanks for asking

aSemy:
POC completed - will be finalised in https://github.com/akuleshov7/ktoml/pull/171

akuleshov7:
no-no, let's not close it! I will merge it to get tests as a separate PR :)

aSemy:
Ahh okay, I misunderstood 👍

ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/decoders/PrimitivesDecoderTest.kt
```diff
@@ -0,0 +1,274 @@
...
+
+            @Serializable
+            data class Data(val value: Byte)
+
+            assertFailsWith<ParseException>(input) {
+                Toml.decodeFromString(Data.serializer(), toml)
+            }
+        }
+
+        testFails("${Char.MAX_VALUE.digitToInt() + 1}")
```
akuleshov7, line 82: `kotlin.IllegalArgumentException: Char ￿ is not a decimal digit`

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
```diff
@@ -26,8 +26,8 @@ public abstract class TomlAbstractDecoder : AbstractDecoder() {
     override fun decodeByte(): Byte = decodePrimitiveType()
     override fun decodeShort(): Short = decodePrimitiveType()
     override fun decodeInt(): Int = decodePrimitiveType()
-    override fun decodeFloat(): Float = decodePrimitiveType()
-    override fun decodeChar(): Char = decodePrimitiveType()
+    override fun decodeFloat(): Float = invalidType("Float", "Double")
```
akuleshov7, line 29: Let's split this logic into several PRs, same with Char.

Actually Char is even more discussible, because there is no syntax in Toml for char literals, only for Strings...

akuleshov7 (approved):
lgtm with my changes :D

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
aSemy: Yeah, I copy and pasted Char in, but in retrospect I don't think it's clear how it should be handled. Is Char a number? Should `101` be treated as ascii and be decoded to `A`? (No, probably not.) What about decoding `a` to `val c: Char`?

PR #165, opened 2022-11-23
Separate parsing from writing in node constructors

### LINKED ISSUE #160: Restructure nodes to avoid parsing in encoders
There's still some remaining parsing code in node constructors making encoding awkward. For example, to create a table, you have to pass a header string to be parsed, rather than the table's key:

```kotlin
val table = TomlTablePrimitive(content = "[$key]", /* ... */)
```

The base `TomlNode` class has a `content` property in its primary constructor which is unused except for exceptions and pretty-printing. For inline tables, I've passed an empty string for this. For other pairs, it's formatted in the constructor. Since it's only used for debugging and exceptions, would it make more sense to create these as needed?

The `lineNo` property also seems to be exclusively used in exceptions.

`TomlNode` also keeps an instance of the input config, used in `insertTableToTree` to construct sub-tables, and `TomlInlineTable.returnTable`. If these functions used a non-parsing constructor, this may be avoidable.

Somewhat related is the usage `Any` in `TomlValue` rather than the actual value type. Differentiation between constructors in the case of strings will be difficult. Maybe that's a separate issue though?

There's probably something I missed

NightEule5:
Fixes #160


NightEule5:
Hmm

```
> Task :ktoml-core:compileTestKotlinMingwX64 FAILED
e: java.lang.OutOfMemoryError: Metaspace
```

akuleshov7:
Small problem is that we need to merge the work related to Long/Int numbers first, let's take it also into the next release

ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/parsers/ArraysOfTablesTest.kt
```diff
@@ -589,13 +589,13 @@ class ArraysOfTablesTest {
                 | - TomlFile (rootNode)
                 |     - TomlArrayOfTables ([[fruit]])
                 |         - TomlArrayOfTablesElement (technical_node)
-                |             - TomlTablePrimitive ([fruit.physical]  )
-                |                 - TomlKeyValuePrimitive (color=red)
-                |                 - TomlKeyValuePrimitive (shape=round)
+                |             - TomlTablePrimitive ([fruit.physical])
+                |                 - TomlKeyValuePrimitive (color="red")
```
akuleshov7, line 593: hm, why we did not have quotes here before? 🤔

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/tables/TomlTablePrimitive.kt
```diff
@@ -17,57 +15,34 @@ import com.akuleshov7.ktoml.writers.TomlEmitter
...
-    // short table name (only the name without parental prefix, like a - it is used in decoder and encoder)
-    override val name: String
-
-    // list of tables (including sub-tables) that are included in this table  (e.g.: {a, a.b, a.b.c} in a.b.c)
-    public override lateinit var tablesList: List<String>
-
-    // full name of the table (like a.b.c.d)
-    public override lateinit var fullTableName: String
-
-    init {
```
akuleshov7, line 47: I see that you moved this part of code to static method in `TomlTable` - is done for some unification or some other ideas under the hood?

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/pairs/keys/TomlKey.kt
```diff
@@ -72,4 +80,27 @@ public class TomlKey(public val rawContent: String, public val lineNo: Int) {
...
     }
+
+    override fun toString(): String = buildString {
+        val emitter = TomlStringEmitter(this, TomlOutputConfig())
+
+        write(emitter)
+    }
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) {
```
akuleshov7, line 91: let's rewrite it in a little bit Kotlin manner? :)
```
return when {
    this == other ->true
    ...
    else -> true
}
```

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/pairs/values/TomlValue.kt
```diff
@@ -10,9 +10,10 @@ import com.akuleshov7.ktoml.writers.TomlEmitter
 
 /**
  * Base class for all nodes that represent values
- * @property lineNo - line number of original file
  */
-public sealed class TomlValue(public val lineNo: Int) {
+public sealed class TomlValue {
+    @Deprecated(message = "lineNo is deprecated. Will be removed in next releases.")
```
akuleshov7, line 15: I will argue on this. I still think that `lineNo` is very important on the side of nodes. 
For debug purposes and for the future usage. In good AST they always store it inside nodes... Actually this is the only real concern in this PR for me.

@NightEule5 why you do not like line number here so much? :D

akuleshov7 (approved):
Lgtm, but the only serious concern I have here is related to `lineNo`

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/pairs/values/TomlValue.kt
akuleshov7: Can I isolate this change in a separate PR?

akuleshov7: after looking more into the code, I understood your pain. 
Yes, we can really remove lineNo from TomlKey and TomlValue: but anyway we should keep it on KeyValue and Table level, it's very important...

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/tables/TomlTablePrimitive.kt
NightEule5: Yes

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/pairs/values/TomlValue.kt
NightEule5: Sure, we can add it back. I just thought it's only used in the parsing process

PR #168, opened 2022-12-08
Fix bug with inline comment

### LINKED ISSUE #167: Bug with " symbol in inline comment
Library version: 0.3.0

Can be reproduced with a test:
```
    @Test
    fun symbolsAfterComment() {
        val keyValue = "test_key = \"test_value\"  # \" some comment".splitKeyValue(0)
        assertEquals("test_value", TomlKeyValuePrimitive(keyValue, 0).value.content)
    }
```
This is a valid TOML format, but I get an exception:

```
Line 0: According to the TOML specification string values (even Enums) should be wrapped (start and end) with quotes (""), but the following value was not: <"test_value"  # " some comment>. Please note that multiline strings are not yet supported.
com.akuleshov7.ktoml.exceptions.ParseException: Line 0: According to the TOML specification string values (even Enums) should be wrapped (start and end) with quotes (""), but the following value was not: <"test_value"  # " some comment>. Please note that multiline strings are not yet supported.
```

Where the problem is: file `TomlKeyValue`, `splitKeyValue` function.

BOOMeranGG:
Fix #167

### DIFF
--- a/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/parsers/StringUtils.kt
+++ b/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/parsers/StringUtils.kt
@@ -93,13 +93,14 @@ internal fun String.trimDoubleBrackets(): String = trimSymbols(this, "[[", "]]")
 /**
  * Takes only the text before a comment
  *
+ * @param allowEscapedQuotesInLiteralStrings value from TomlInputConfig
  * @return The text before a comment, i.e.
  * ```kotlin
  * "a = 0 # Comment".takeBeforeComment() == "a = 0"
  * ```
  */
-internal fun String.takeBeforeComment(): String {
-    val commentStartIndex = getCommentStartIndex()
+internal fun String.takeBeforeComment(allowEscapedQuotesInLiteralStrings: Boolean): String {
+    val commentStartIndex = getCommentStartIndex(allowEscapedQuotesInLiteralStrings)
 
     return if (commentStartIndex == -1) {
         this.trim()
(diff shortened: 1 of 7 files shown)


ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/parsers/StringUtils.kt
```diff
@@ -167,6 +173,33 @@ private fun String.validateSymbols(lineNo: Int) {
     }
 }
 
+private fun String.getCommentStartIndex(): Int {
+    // temporary replace \" and \' to hide non-significant quotes
+    // we cannot remove them because we need to know the original index
+    val chars = this
+        .replace("\\\"", "__")
+        .replace("\\\'", "__")
```
NightEule5, line 181: The TOML spec allows backslashes to be interpreted literally in single-quoted strings, we just have a config option to enable single quote escaping. Would this work with, say, `a = 'C:\some\path\'`?

BOOMeranGG: Good point! It doesn't work for this test

BOOMeranGG: Fixed. You can see additional tests in StringUtilsTest, now they work

BOOMeranGG:
@akuleshov7 Previous time one check failed. Is something wrong?

akuleshov7:
> @akuleshov7 Previous time one check failed. Is something wrong?
approved run, let me check. Have you run `./gradlew diktatFixAll` or `./gradlew detekt` tasks?

akuleshov7 (approved):

BOOMeranGG:
> Have you run ./gradlew diktatFixAll or ./gradlew detekt tasks?
Yes, I have. It's all right. Let's merge?

akuleshov7:
> Have you run ./gradlew diktatFixAll or ./gradlew detekt tasks?
> Yes, I have. It's all right. Let's merge?

Great, thank you so much for this fix!

PR #171, opened 2022-12-27
Limitation for Long type finally removed: support for integer types

akuleshov7:
### What's done:
- removed limitation for integers to be long
- TOML spec: Arbitrary 64-bit signed integers (from −2^63 to 2^63−1) should be accepted and handled losslessly. If an integer cannot be represented losslessly, an error must be thrown
- so right now we support Byte, Short and Int. But if there is an overflow, we throw an exception
- BREAKING CHANGE: `CastException` removed and now unified with `IllegalTypeException`

### DIFF
--- a/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
+++ b/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
@@ -4,6 +4,7 @@ import com.akuleshov7.ktoml.exceptions.IllegalTypeException
 import com.akuleshov7.ktoml.tree.nodes.TomlKeyValue
 import com.akuleshov7.ktoml.tree.nodes.pairs.values.TomlLong
 import com.akuleshov7.ktoml.utils.IntegerLimitsEnum
+import com.akuleshov7.ktoml.utils.IntegerLimitsEnum.*
 import kotlinx.datetime.Instant
 import kotlinx.datetime.LocalDate
 import kotlinx.datetime.LocalDateTime
@@ -79,14 +80,12 @@ public abstract class TomlAbstractDecoder : AbstractDecoder() {
      * This method does simple validation of integer values to avoid overflow. For example, you really want to use byte,
      * we will check here, that your byte value does not exceed 127 and so on.
      */
-    private inline fun <reified T> decodeInteger(content: Long, lineNo: Int): T {
-        return when (T::class) {
-            Byte::class -> validateAndConvertInt(content, lineNo, IntegerLimitsEnum.BYTE) { c: Long -> c.toByte() as T }
-            Short::class -> validateAndConvertInt(content, lineNo, IntegerLimitsEnum.SHORT) { c: Long -> c.toShort() as T  }
-            Int::class -> validateAndConvertInt(content, lineNo, IntegerLimitsEnum.INT) { c: Long -> c.toInt() as T  }
-            Long::class -> validateAndConvertInt(content, lineNo, IntegerLimitsEnum.LONG) { c: Long -> c as T  }
-            else -> invalidType(T::class.toString(), "Signed Type")
-        }
+    private inline fun <reified T> decodeInteger(content: Long, lineNo: Int): T = when (T::class) {
+        Byte::class -> validateAndConvertInt(content, lineNo, BYTE) { num: Long -> num.toByte() as T }
+        Short::class -> validateAndConvertInt(content, lineNo, SHORT) { num: Long -> num.toShort() as T }
+        Int::class -> validateAndConvertInt(content, lineNo, INT) { num: Long -> num.toInt() as T }
+        Long::class -> validateAndConvertInt(content, lineNo, LONG) { num: Long -> num as T }
+        else -> invalidType(T::class.toString(), "Signed Type")
     }
 
     private inline fun <reified T> validateAndConvertInt(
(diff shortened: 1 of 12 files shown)


akuleshov7:
@aSemy @NightEule5 @BOOMeranGG @nulls could you please have a look?

nulls (approved):

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/utils/Types.kt
```diff
@@ -0,0 +1,35 @@
...
+ */
+public enum class IntegerLimitsEnum(public val min: Long, public val max: Long) {
+    BYTE(Byte.MIN_VALUE.toLong(), Byte.MAX_VALUE.toLong()),
+    INT(Int.MIN_VALUE.toLong(), Int.MAX_VALUE.toLong()),
+    LONG(Long.MIN_VALUE, Long.MAX_VALUE),
+    SHORT(Short.MIN_VALUE.toLong(), Short.MAX_VALUE.toLong()),
+    ;
+    // Unsigned values are not supported now, and I think
+    // that will not be supported, because TOML spec says the following:
+    // Arbitrary 64-bit signed integers (from −2^63 to 2^63−1) should be accepted and handled losslessly.
```
aSemy, line 19: Would you mind elaborating on why unsigneds can't be supported?

I'd recommend making a new GitHub issue 'support unsigned ints' (I could make the issue if you like) and adding the link in the source code, then it's easier to track and follow progress.

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
```diff
@@ -49,25 +51,62 @@ public abstract class TomlAbstractDecoder : AbstractDecoder() {
...
+    }
+
+    private inline fun <reified T> validateAndConvertInt(
+        content: Long,
+        lineNo: Int,
+        limits: IntegerLimitsEnum,
+        conversion: (Long) -> T,
+    ): T = if (content in limits.min..limits.max) {
+        conversion(content)
+    } else {
```
aSemy, line 98: I think this code is based on my PR which used the MIN/MAX boundaries, but I checked the Kotlinx Serialization JSON decoder and I think it has a better way.

https://github.com/Kotlin/kotlinx.serialization/blob/d7e58c2eeb02e1da87ad36d41ef25523fafb6935/formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt#L299-L304

1. decode to Long
2. convert to `value.toInt()`
3. if the result is different, then throw an exception

It's less code, and I think easier to understand

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
```diff
@@ -49,25 +51,62 @@ public abstract class TomlAbstractDecoder : AbstractDecoder() {
...
+            throw IllegalTypeException(
                 "Cannot decode the key [${keyValue.key.content}] with the value [${keyValue.value.content}]" +
                         " with the provided type [${T::class}]. Please check the type in your Serializable class or it's nullability",
                 keyValue.lineNo
             )
         }
     }
+
+    /**
+     * After a lot of discussions, we have finally decided to allow to use Integer types and not only Long.
```
aSemy, line 79: > After a lot of discussions

I think it would be valuable to add a link to the discussions, so there's some context

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/utils/Types.kt
akuleshov7: No, I don't mind, but the main problem is related to the kotlinx.serialization. I cannot find decodeUInt in their library...

aSemy: Yeah, it's a little bit odd how KxS handles unsigneds. 

KxS JSON handles them natively though https://github.com/Kotlin/kotlinx.serialization/blob/v1.4.1/docs/value-classes.md#unsigned-types-support-json-only

I suspect that they should 'just work' though, since they have serializers https://github.com/Kotlin/kotlinx.serialization/blob/v1.4.1/core/commonMain/src/kotlinx/serialization/internal/ValueClasses.kt

How KxS JSON handles them is

1. unsigned numbers are value classes, so their `SerialDescriptor`s are 'inline'
2. it keeps a list of the SerialDescriptors for all unsigned numbers
   https://github.com/Kotlin/kotlinx.serialization/blob/dc950f5098162a3f8dd742d04b95f9a0405470e1/formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonEncoder.kt#L15-L24
4. there's a specific function [`AbstractDecoder.decodeInline(...)`](https://github.com/Kotlin/kotlinx.serialization/blob/dc950f5098162a3f8dd742d04b95f9a0405470e1/core/commonMain/src/kotlinx/serialization/encoding/AbstractDecoder.kt#LL37C39-L37C39), that will handle the unsigned numbers
5. then there's a specific decoder for the unsigned types https://github.com/Kotlin/kotlinx.serialization/blob/d7e58c2eeb02e1da87ad36d41ef25523fafb6935/formats/json/commonMain/src/kotlinx/serialization/json/internal/StreamingJsonDecoder.kt#L367-L379

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
akuleshov7: The only problem I see here is that comparing operation with limits can be easier than making a `toInt()` and than comparing... But anyway probably your variant is good enough.

akuleshov7: speaking about the idea - I actually asked a question on stack overflow about that and as well as I and you thought - everyone suggests to check boundaries: https://stackoverflow.com/questions/74510016/how-to-understand-if-there-was-an-integer-overflow-during-parsing-of-string-with :D 

So probably, it is a good idea to have the most obvious solution

aSemy: Personally I prefer the KxS JSON way with `.toInt()`, but the result is the same pick whatever you prefer :)

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/utils/Types.kt
akuleshov7: We will need to add it then also... Yes, let's support it then :)
Looks like a workaround in KxS, looks like they forgot about unsigned in the very beginning

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
akuleshov7: Also funny thing that @NightEule5 suggested same solution as you here: https://github.com/akuleshov7/ktoml/pull/153#discussion_r1003881958 :D

PR #175, opened 2022-12-30
Removing a Double restriction: float support

akuleshov7:
### What's done:
- enabled the decoding of Float type

### DIFF
--- a/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
+++ b/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
@@ -39,19 +39,19 @@ public abstract class TomlAbstractDecoder : AbstractDecoder() {
     override fun decodeString(): String = decodePrimitiveType()
 
     protected fun DeserializationStrategy<*>.isDateTime(): Boolean =
-        descriptor == instantSerializer.descriptor ||
-                descriptor == localDateTimeSerializer.descriptor ||
-                descriptor == localDateSerializer.descriptor
+            descriptor == instantSerializer.descriptor ||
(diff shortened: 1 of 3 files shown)


ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
```diff
@@ -75,36 +80,70 @@ public abstract class TomlAbstractDecoder : AbstractDecoder() {
...
+    /**
+     * ktoml parser treats all integer literals as Long and all floating-point literals as Double,
+     * so here we should be checking that there is no overflow with smaller types like Byte, Short and Int.
+     */
+    private inline fun <reified T> validateAndConvertFloatingPoint(
+        content: Double,
+        lineNo: Int,
+        limits: FloatingPointLimitsEnum,
+        conversion: (Double) -> T,
+    ): T = if (content in limits.min..limits.max) {
```
nulls, line 98: what's about suggestion to convert to lower type and then back and check that value is the same?

akuleshov7: @nulls , such suggestions valid, however looks a little bit confusing for integer types. But for floating-point types it can cause an error. 

Floating-point arithmetics in Kotlin compiler is partially implemented using C++ interop. And in C++ standard it is explicitly told that floating point numbers cannot store all values exactly (not due to limits, but due to precision), so there can potentially be cases when `doubleNumber.toFloat().toDouble() != doubleNumber` due to loss of bits. 

So it is a big question for me if I really should rely on such hack and have potential (but with low probability) case when we corrupt the number and throw an exception with a higher complexity instead of simple check of bounds.

As you know, I am not a risky guy, so I think that I would better use low risk option here (at least with floating point numbers).

PR #176, opened 2023-01-03
Adding the support for Char

akuleshov7:
### What's done:
- conversion from string to char
- enabling some old tests
- still we need unicode conversion

### DIFF
--- a/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
+++ b/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
@@ -2,7 +2,9 @@ package com.akuleshov7.ktoml.decoders
 
 import com.akuleshov7.ktoml.exceptions.IllegalTypeException
 import com.akuleshov7.ktoml.tree.nodes.TomlKeyValue
+import com.akuleshov7.ktoml.tree.nodes.pairs.values.TomlBasicString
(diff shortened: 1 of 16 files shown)


ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
```diff
@@ -32,7 +34,35 @@ public abstract class TomlAbstractDecoder : AbstractDecoder() {
     override fun decodeShort(): Short = decodePrimitiveType()
     override fun decodeInt(): Int = decodePrimitiveType()
     override fun decodeFloat(): Float = decodePrimitiveType()
-    override fun decodeChar(): Char = invalidType("Char", "String")
+    override fun decodeChar(): Char {
+        val keyValue = decodeKeyValue()
+        return when (val value = keyValue.value) {
```
akuleshov7, line 39: We will give users a possibility to decode Char from integer or from a literal

akuleshov7:
@aSemy @nulls @NightEule5 @BOOMeranGG I hope you don't mind if I will merge it without review, anyway we will polish it later...

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
akuleshov7: char1 = 156
char2 = '5'

TOML standard does not support Char, but we will have it out of the box

PR #182, opened 2023-01-27
Lower JVM target to 1.8

saschpe:
Allows to use ktoml (at runtime) on older JVMs. Since Kotlin currently targets 1.8 IL, there's zero difference at compile or test time.

### DIFF
--- a/ktoml-core/build.gradle.kts
+++ b/ktoml-core/build.gradle.kts
@@ -19,9 +19,7 @@ kotlin {
     // building jvm task only on windows
     jvm {
         compilations.all {
-            kotlinOptions {
-                jvmTarget = "11"
-            }
+            kotlinOptions.jvmTarget = "1.8"
         }
     }
 
--- a/ktoml-file/build.gradle.kts
+++ b/ktoml-file/build.gradle.kts
@@ -11,9 +11,7 @@ kotlin {
 
     jvm {
         compilations.all {
-            kotlinOptions {
-                jvmTarget = "11"
-            }
+            kotlinOptions.jvmTarget = "1.8"
         }
     }
 


akuleshov7:
Hey, thank you for a contribution!
I think there can be some miss-compatibilities between Java bytecode, but definitely 1.11 -> 1.8 should be ok for us. 

But the release with this change will not be very soon, do you need it urgently?

akuleshov7:
But usually that is a little bit weird to make a downgrade in a version 😄

akuleshov7 (approved):

saschpe:
Yeah, it's for a proprietary product ;-) I don't think there will be any bytecode incompatibility since Kotlin targets JVM-1.8 since Kotlin-1.5.x. Even the recent Kotlin-1.8 release only started to compile their sodlib against JVM-1.8 language level. 

btw.: The CI job issue seems to be a crashed Gradle daemon on that Ubuntu VM.

NightEule5:
One thing I'd point out with downgrading is that since 1.5.20, on Java 9+ string concatenation and templates is optimized with [`invokedynamic`](https://kotlinlang.org/docs/whatsnew1520.html#string-concatenation-via-invokedynamic). Targeting 1.8 bytecode generates the old implementation with `StringBuilder`. This may not be significantly slower, but it's something to keep in mind

saschpe:
Good point, though the added benefit of a wider range of JVMs that can be used is a good think. Thanks for that quick turnaround ;-)

saschpe:
> But the release with this change will not be very soon, do you need it urgently?

Kind of. Would it be a big hassle to do a 0.4.1 release? TIA!

akuleshov7:
> > But the release with this change will not be very soon, do you need it urgently?
> 
> Kind of. Would it be a big hassle to do a 0.4.1 release? TIA!

I am in a very good mood today 😄 , 0.4.1 is releasing :)

saschpe:
Looks like I gonna have to return the favor by fixing something? Thanks ;-)

akuleshov7:
it looks like Java guys have some issues with that: https://github.com/akuleshov7/ktoml/issues/191

PR #183, opened 2023-01-28
Support string and array multiline, and integer representation encoding

### LINKED ISSUE #161: Support TomlMultiline and TomlInteger annotations
Currently, when `TomlMultiline` and `TomlInteger` are encountered during serialization, an `UnsupportedEncodingFeatureException` is thrown. This is because the tree doesn't support these yet. `TomlMultiline` would apply to basic and literal string pairs, and possibly primitive arrays. `TomlInteger` would apply to integer pairs.

We could add a `multiline` property to the primary constructor of `TomlLiteralString`, `TomlBasicString`, and `TomlArray`, like this:

```kotlin
public class TomlLiteralString
internal constructor(
    override var content: Any,
    lineNo: Int,
    public val multiline: Boolean,
)
```

And a `representation` property to `TomlLong`:

```kotlin
public class TomlLong
internal constructor(
    override var content: Any,
    lineNo: Int,
    public val representation: IntegerRepresentation
)
```

NightEule5:
Fixes #161


BOOMeranGG:
One more suggestion is: 
Do the same with grouping as in https://kotlinlang.org/docs/numbers.html#literal-constants-for-numbers (to be able to use Grouped along with BINARY/OCTAL/HEX)

To get the following result:
```
hexBytes = 0xFF_EC_DE_5E
bytes = 0b11010010_01101001_10010100_10010010
```

What do you think?

NightEule5:
Yep, I thought about that and it's definitely doable, but I figured that would be a feature to add later. We can make an issue about this.

There's two ways I could see this being represented in the `TomlInteger` annotation. We could simply remove the `GROUPED` enum value and use a `groupSize` property instead, `0` being no grouping. We could make a separate enum, like `IntegerGrouping`, with [common groupings](https://en.wikipedia.org/wiki/Decimal_separator#Digit_grouping) like `THOUSANDS` or `BYTES`.

I think the first would be preferred

BOOMeranGG:
> We could simply remove the GROUPED enum value and use a groupSize property instead, 0 being no grouping.

If groupSize = 3 what does it mean? Will it be like this: `1_000_000` ?
Also for the first approach we could add constants for common groupings

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/writers/TomlEmitter.kt
```diff
@@ -210,7 +210,25 @@ public abstract class TomlEmitter(config: TomlOutputConfig) {
                 OCTAL ->
                     emit("0o")
                         .emit(integer.toString(8))
-                GROUPED -> TODO()
+                GROUPED ->
+                    emit(
+                        if (integer < 1000) {
+                            // No grouping needed.
+                            integer.toString()
```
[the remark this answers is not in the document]
NightEule5: Good catch, I'll add that

NightEule5:
> If groupSize = 3 what does it mean? Will it be like this: `1_000_000` ? Also for the first approach we could add constants for common groupings

Yep, that's the idea

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/writers/TomlEmitter.kt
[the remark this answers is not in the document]
NightEule5: I realized I hadn't handled the negative case for any of the non-decimal cased too. Fixed

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/writers/TomlEmitter.kt
```diff
@@ -198,20 +197,31 @@ public abstract class TomlEmitter(config: TomlOutputConfig) {
...
      * @return this instance
      */
-    public fun emitValue(integer: Long, representation: IntegerRepresentation = DECIMAL): TomlEmitter =
+    @Suppress("MAGIC_NUMBER")
+    public fun emitValue(integer: Long, representation: IntegerRepresentation = DECIMAL): TomlEmitter {
+        var value = integer
+
+        if (value < 0) {
+            return emit('-').emitValue(-value, representation)
+        }
```
BOOMeranGG, line 206: If we have Long.MIN_VALUE we get a StackOverflow error because 
`-Long.MIN_VALUE == Long.MIN_VALUE`

BOOMeranGG (changes requested):

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/writers/TomlEmitter.kt
BOOMeranGG: Probably we could convert number to string first and then add grouping. Or any other idea :)

NightEule5: Not giving up on a math-based solution just yet :P

BOOMeranGG: I just think, if we'll add `groupSize` in the future, is it easier to work with string?

NightEule5: Maybe

BOOMeranGG (approved):

NightEule5:
Surprisingly not much conflicts

akuleshov7:
> Surprisingly not much conflicts

yeah, I am also confused - came here to look at conflicts :)
Will review and merge it today

akuleshov7:
It will also be good to add a description with all such annotations like `@TomlMultiline` to readme. But can be done in a separate issue. Otherwise, I'm afraid, we will get a lot of questions about that

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/pairs/values/TomlLong.kt
```diff
@@ -1,22 +1,53 @@
...
  * @property content
+ * @property representation The representation of the integer.
  */
 public class TomlLong internal constructor(
-    override var content: Any
+    override var content: Any,
+    public var representation: IntegerRepresentation = DECIMAL
 ) : TomlValue() {
-    public constructor(content: String, lineNo: Int) : this(content.toLong())
+    public constructor(content: String, lineNo: Int) : this(content.parse())
```
akuleshov7, line 17: Actually I do not get it - if you remember I was always avoiding custom logic for parsing and even made logic for parsing based on exceptions: 
```
try {
    s.toLong
} catch() {
    try {
        s.toInt
    }
}

...
etc.
```

So I tried as much as possible to avoid custom parsing. Why do we need to have it here?

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/pairs/values/TomlLiteralString.kt
```diff
@@ -92,8 +88,25 @@ public class TomlLiteralString internal constructor(
          */
         private fun String.convertSingleQuotes(): String = this.replace("\\'", "'")
 
-        private fun String.escapeQuotesAndVerify(config: TomlOutputConfig) =
+        private fun String.escapeQuotesAndVerify(config: TomlOutputConfig, multiline: Boolean) =
                 when {
+                    multiline ->
+                        when {
+                            multilineControlCharacterRegex in this ->
```
akuleshov7, line 95: Regular expressions affect performance and in some cases can lead to exceptions in case of very long strings...
May it will be better to use something like `string.any(setOfillegalCharacters::contains)` instead?

akuleshov7 (approved):
Basically lgtm, but I am worried about:
1) custom parsing of Long
2) a little bit unreadable regular expressions that can also affect the performance

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/pairs/values/TomlLong.kt
NightEule5: Hmm, I did it this way to support parsing the other integer types, instead of just decimal. `toLong` only parses decimal numbers, it can't handle the the prefixes and non-decimal bases that TOML supports, like `0b`, `0x`, etc. We can do it a different way if you have something in mind

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/pairs/values/TomlLiteralString.kt
NightEule5: Sure, admittedly regexes aren't needed here. I assume you want a separate issue for this?

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/pairs/values/TomlLong.kt
akuleshov7: I thought that this parsing of prefixes is included into `toLong`… I got the intention now

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/pairs/values/TomlLiteralString.kt
akuleshov7: Yeah, I will make a separate one

PR #184, opened 2023-01-28
Support parsing for multiline string

### LINKED ISSUE #46: Multiline literal strings
Multi-line literal strings are surrounded by three single quotes on each side and allow newlines. Like literal strings, there is no escaping whatsoever. A newline immediately following the opening delimiter will be trimmed. All other content between the delimiters is interpreted as-is without modification.
```
regex2 = '''I [dw]on't need \d{2} apples'''
lines  = '''
The first newline is
trimmed in raw strings.
   All other whitespace
   is preserved.
'''
```

BOOMeranGG:
Fixes #46


akuleshov7:
@NightEule5 you will get conflicts with this one PR, I guess… 🥲

NightEule5:
> @NightEule5 you will get conflicts with this one PR, I guess… smiling_face_with_tear

Ok with me

ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/decoders/multiline/BasicMultilineStringDecoderTest.kt
```diff
@@ -0,0 +1,236 @@
...
+            second line \
+            
+            third line""${'"'}
+        """.trimIndent()
+        assertEquals(SimpleString("first line second line third line"), Toml.decodeFromString(test))
+
+        test = "a = \"\"\"\n" +
+            "The quick brown \\\n\n\n" +
+            "  fox jumps over \\\n" +
+            "    the lazy dog.\"\"\""
```
NightEule5, line 93: Why not use multiline strings here? If it's the quotes, when I have to write triple quotes I usually use templates similar to what you did above this. The strings here are bit hard to read

Something like this should work:

```kotlin
val tripleQuotes = "\"\"\""

test = """
a = $tripleQuotes
The quick brown \



  fox jumps over \
    the lazy dog.$tripleQuotes
""".trimIndex()
```

NightEule5:
Nothing else to add, looks good to me. We'll see what akuleshov7 thinks, parsing is more his thing

ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/decoders/multiline/BasicMultilineStringDecoderTest.kt
BOOMeranGG: Added

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/pairs/TomlKeyValue.kt
```diff
@@ -138,13 +138,7 @@ public fun String.parseValue(lineNo: Int, config: TomlInputConfig): TomlValue =
...
     // ===== strings
     else -> when (this.first()) {
-        '\'' -> if (this.startsWith("'''")) {
-            // actually the support for multiline strings will be here, but now just a simple mock
-            // this mock will raise an exception here
-            TomlBasicString(this, lineNo)
-        } else {
-            TomlLiteralString(this, lineNo, config)
-        }
+        '\'' -> TomlLiteralString(this, lineNo, config)
```
akuleshov7, line 141: So we received a multiline string and are parsing it with `TomlLiteralString`? 
It's ok, if it's not a mistake, but while reading the code - I would expect here different types of strings that start from `'`: Literal, Multiline, etc. 

As this code was intended to be an entry point for determination of different types.

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/pairs/TomlKeyValue.kt
```diff
@@ -138,13 +138,7 @@ public fun String.parseValue(lineNo: Int, config: TomlInputConfig): TomlValue =
...
         // ===== basic strings
         '\"' -> TomlBasicString(this, lineNo)
```
akuleshov7, line 143: So we received a multiline string and are parsing it with `TomlBasicString`? 
It's ok, if it's not a mistake, but while reading the code - I would expect here different types of strings that start from `"`: Basic, Multiline, etc. 

As this code was intended to be an entry point for determination of different types. So I got a little bit confused.

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/pairs/values/TomlBasicString.kt
```diff
@@ -43,19 +46,25 @@ public class TomlBasicString internal constructor(
     public companion object {
         private fun String.verifyAndTrimQuotes(lineNo: Int): Any =
                 when {
+                    // ====== multiline string (""") =======
```
akuleshov7, line 49: Yeah, I see: you decided not to make differences on the higher level. 
That's actually okay, may I should have done the same with LiteralStrings

akuleshov7:
@BOOMeranGG I think we have missed multiline literal strings also:

image

Can be done separately, np

akuleshov7 (approved):
Looked more closely, actually awesome work, lgtm!
Thank you.

PS great work with tests also

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/parsers/TomlParser.kt
```diff
@@ -135,52 +128,91 @@ public value class TomlParser(private val config: TomlInputConfig) {
...
         val firstEqualsSign = line.indexOfFirst { it == '=' }
         if (firstEqualsSign == -1) {
-            return false
+            return MultilineType.NOT_A_MULTILINE
         }
         val value = line.substring(firstEqualsSign + 1).trim()
 
-        return value.startsWith("[") &&
-                value.endsWith("]").not()
+        if (value.startsWith("[") && !value.endsWith("]")) {
```
akuleshov7, line 188: the only stylistic comment I have here - is usage of `return when { 

}
`

It usually looks much better with a huge number of `if statements`

PR #190, opened 2023-03-12
Replace regexes with functions

### LINKED ISSUE #187: Remove complex regular expressions and replace it with simple validation
Regular expressions affect performance and in some cases can lead to exceptions in case of very long strings...
May it will be better to use something like `string.any(setOfillegalCharacters::contains)` instead?

_Originally posted by @akuleshov7 in https://github.com/akuleshov7/ktoml/pull/183#discussion_r1113227164_

NightEule5:
Fixes #187

Tests pass, but not sure if there are edge cases I didn't account for


NightEule5:
The build timed out???

akuleshov7:
> The build timed out???
> 
> Edit: Now it's OutOfMemoryException :/

yeah, it's looks to be Kotlin error. Will ask JB guys about this. But the issue is very flaky - looks like some leaks in memory

akuleshov7:
Updated branch and restarted the build

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/utils/Keys.kt
```diff
@@ -0,0 +1,37 @@
...
+ * alphanumeric, underscore, and hyphen characters.
+ *
+ * @return `true` if this key is a bare key.
+ */
+internal fun String.isBareKey() =
+        all {
+            it in 'A'..'Z' ||
+                    it in 'a'..'z' ||
+                    it in '0'..'9' ||
+                    it in "-_"
```
akuleshov7, line 18: May be better to make a set of symbols instead of several OR (||) checks:
`it in (('A'..'Z') + ('a'..'z') + ('0'..'9') + "-_")`

akuleshov7, line 18: Btw - why "-_" here? Why two symbols? I thought that we need to check only underscore

NightEule5: Well, in the regex we did it too. A bare key can contain a hyphen

```kotlin
internal val bareKeyRegex = Regex("[A-Za-z0-9_-]+")
```

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/utils/SpecialCharacters.kt
```diff
@@ -101,40 +101,90 @@ public fun StringBuilder.appendEscapedUnicode(
  */
 public fun String.escapeSpecialCharacters(multiline: Boolean = false): String =
         if (multiline) {
-            val withCtrlCharsEscaped = escapeCtrlChars(multilineControlCharacterRegex)
-            val withQuotesEscaped = withCtrlCharsEscaped.replace("\"\"\"", "\"\"\\\"")
```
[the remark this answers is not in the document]
NightEule5: We could escape all of them, but there's not much of a reason to in multiline strings. Only escaping the last of the three is a pattern I saw in the spec, so I went with that:

```toml
str4 = """Here are two quotation marks: "". Simple enough."""
# str5 = """Here are three quotation marks: """."""  # INVALID
str5 = """Here are three quotation marks: ""\"."""
str6 = """Here are fifteen quotation marks: ""\"""\"""\"""\"""\"."""

# "This," she said, "is just a pointless statement."
str7 = """"This," she said, "is just a pointless statement.""""
```

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/utils/Keys.kt
NightEule5: Good point. Could even add a toString at the end and make it a static property

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/utils/SpecialCharacters.kt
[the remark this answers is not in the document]
akuleshov7: ah, got it, thank you

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/utils/Keys.kt
akuleshov7: ok, got it

akuleshov7 (approved):
lgtm, thank you so much!

PR #217, opened 2023-05-05
Adding a test to validate line numbers and fix for multiline strings

akuleshov7:
### What's done:
- On multiline strings we forgot to update isMultiline flag while decoding and parsing
- Adding extra info to prettyPrint
- Adding a test for validation of line numbers
- Removing deprecated 'content' from the TomlNode


ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/parsers/SetLineNoTest.kt
```diff
@@ -0,0 +1,58 @@
...
+        parsedToml.prettyPrint(true)
+        assertEquals(
+            """
+                 | - TomlFile (rootNode)[line:0]
+                 |     - TomlTablePrimitive ([a])[line:4]
+                 |         - TomlKeyValuePrimitive (test=1)[line:6]
+                 |         - TomlArrayOfTables ([[a.b]])[line:10]
+                 |             - TomlArrayOfTablesElement (technical_node)[line:10]
+                 |                 - TomlKeyValuePrimitive (test=1)[line:11]
+                 |                 - TomlKeyValuePrimitive (mls='''      1
```
akuleshov7, line 48: that's a little bit weird, cause it looks like we forget to put a newline after multiline string delimiters

ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/parsers/SetLineNoTest.kt
```diff
@@ -0,0 +1,58 @@
...
+             mls = '''
+                  1
+                  2
+                  3
+             '''
+             
+             mla = [
+                 "a",
+                 "b",
+                 "c"
```
akuleshov7, line 34: need also to add a test for basic multiline strings:
```
             mls = """
                  1
                  2
                  3
             """
```

ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/parsers/SetLineNoTest.kt
akuleshov7: @NightEule5 while we encode the string, I think it is necessary for the user to have a proper formatting of multiline strings. Now we have the following: 
```
a = """1
2
3"""
```

I think (even the TOML spec is confusing [https://toml.io/en/v1.0.0#string](here)) that we should add a newline there (it will be more convenient):
```
a = """
1
2
3
"""
```

ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/parsers/SetLineNoTest.kt
akuleshov7: added

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/writers/TomlEmitter.kt
```diff
@@ -180,7 +180,9 @@ public abstract class TomlEmitter(config: TomlOutputConfig) {
             val quotes = if (isLiteral) "'''" else "\"\"\""
 
             emit(quotes)
+                .emitNewLine()
```
akuleshov7, line 183: @NightEule5 can you please review it, if possible?

NightEule5: Looks ok

ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/parsers/SetLineNoTest.kt
NightEule5: Yep, true

akuleshov7: And of course I have forgotten to trim spaces before the quotes `'''` while decoding 😢 
https://github.com/akuleshov7/ktoml/issues/221

PR #240, opened 2023-09-02
Add support for empty inline table

### LINKED ISSUE #236: Support for zero-item inline tables
`TomInlineTable` in `ktoml-core-jvm:0.5.0` throw a `ParseException` when it encounters a zero-item inline table.

For example, consider the following line of valid (per [language docs](https://github.com/toml-lang/toml/blob/main/toml.md#inline-table)) TOML:

```
ACTIVITY_COMPOSE = {}
```

This yields the following exception:

```
ParseException: <...>: Incorrect format of Key-Value pair (missing equals sign). Should be <key = value>, but was: . If you wanted to define table - use brackets []
```

BOOMeranGG:
Fix #236

### DIFF
--- a/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/tables/TomlInlineTable.kt
+++ b/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/tables/TomlInlineTable.kt
@@ -131,6 +131,9 @@ public class TomlInlineTable internal constructor(
                 .trimCurlyBraces()
                 .trim()
                 .also {
+                    if (it.isEmpty()) {
+                        return listOf(TomlStubEmptyNode(lineNo))
+                    }
                     if (it.endsWith(",")) {
                         throw ParseException(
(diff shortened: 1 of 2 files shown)


ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/tables/TomlInlineTable.kt
```diff
@@ -131,6 +131,9 @@ public class TomlInlineTable internal constructor(
                 .trimCurlyBraces()
                 .trim()
                 .also {
+                    if (it.isEmpty()) {
```
akuleshov7, line 134: too easy :D

akuleshov7 (approved):

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/tables/TomlInlineTable.kt
BOOMeranGG: I was surprised too

Issue #123, opened 2022-03-14
Switch the testing framework to Kotest?


NightEule5:
Currently kotlin.test is used for testing. Have you considered [Kotest](https://kotest.io) instead? Like Kotlin's test framework, it's built on JUnit on the JVM, but tests are written in a more idiomatic way. There's also an IntelliJ plugin that can display the tests being run in real time, rather than producing a report at the end.

Here's an example of one style it provides:

```kotlin
object ExampleTest : StringSpec({
    "Test string length should be 11" {
        test().shouldHaveLength(11)
    }
})

// ...

fun test() = "test string"
```

This is, in my opinion, way easier to read than:

```kotlin
class ExampleTest {
    @Test
    fun testStringLength() {
        assertEquals(
            test().length,
            11
        )
    }
}

// ...

fun test() = "test string"
```

Transitioning would take some effort, but I think it's worth considering. Thoughts?

akuleshov7:
@bishiboosh @petertrr what do you think about kotest?

sksamuel:
> Good idea, we can try it. But actually you should know, that several people in JetBrains do not like this framework :) 

@akuleshov7 I would be interested to know why some people in Jetbrains do not like the framework? :)

NightEule5:
> I would be interested to know why some people in Jetbrains do not like the framework? :)

As do I. It's great

NightEule5:
> Unfortunately I am busy right now with the support for Custom deserialisers and Inline (VALUE) classes, so won't be able to make it in the nearest future... :(

Understandable :)

petertrr:
> @bishiboosh @petertrr what do you think about kotest?

I haven't tried kotest myself just yet, but I've read about it. Indeed, API looks nice and more kotlin-idiomatic. I was a little bit concerned about multiplatform support and IDE support, though.

As they say in [kotest docs](https://kotest.io/docs/quickstart#test-framework), IDE support for multiplafrom is limited and tests can't be run from IDE directly (though, it usualy doesn't work with kotlin.test too :| ).

I think it's worth a shot, and their library for property-based testing can be useful for testing ktoml, where there are a lot of similar test scenarios for different input data.

PR #243, opened 2023-10-18
Added support for Kotest tests

ILikeYourHat:
Addresses #123 

* Added Kotest framework to project
* Rewritten all tests in `ktoml-file` and `ktoml-source` modules
* Rewritten a few tests in `ktoml-core` module
* Sometimes making minor improvements in tests, like migrating away from `TomlTable.fullTableName` method or removing unused classes

Hope that you would like the way tests look now, and accept my humble contribution :)

### DIFF
--- a/buildSrc/src/main/kotlin/com/akuleshov7/Versions.kt
+++ b/buildSrc/src/main/kotlin/com/akuleshov7/Versions.kt
@@ -5,7 +5,6 @@
 
 object Versions {
     const val KOTLIN = "1.9.21"
-    const val JUNIT = "5.7.1"
     const val OKIO = "3.1.0"
     const val SERIALIZATION = "1.5.0"
     const val KOTEST = "5.7.2"
--- a/buildSrc/src/main/kotlin/com/akuleshov7/Versions.kt
+++ b/buildSrc/src/main/kotlin/com/akuleshov7/Versions.kt
@@ -8,4 +8,5 @@ object Versions {
     const val JUNIT = "5.7.1"
     const val OKIO = "3.1.0"
     const val SERIALIZATION = "1.5.0"
+    const val KOTEST = "5.7.2"
 }
--- a/ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/decoders/PartialDecoderTest.kt
+++ b/ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/decoders/PartialDecoderTest.kt
@@ -1,33 +1,26 @@
 package com.akuleshov7.ktoml.decoders
 
 import com.akuleshov7.ktoml.Toml
+import io.kotest.matchers.shouldBe
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlin.test.Test
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.serializer
-import kotlin.test.assertEquals
 
 @ExperimentalSerializationApi
 class PartialDecoderTest {
-    @Serializable
-    data class TwoTomlTables(val table1: Table1, val table2: Table2)
 
     @Serializable
     data class Table1(val a: Long, val b: Long)
 
-    @Serializable
-    data class Table2(val c: Long, val e: Long, val d: Long)
-
     @Test
     fun testPartialDecoding() {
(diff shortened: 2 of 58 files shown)


bgalek (approved):

akuleshov7:
Hey, thank you, you are awesome, we had big discussions about kotest https://github.com/akuleshov7/ktoml/issues/123

ILikeYourHat:
One more thing: I did this PR because I'm participating in the [Hacktoberfest](https://hacktoberfest.com/). I would be grateful if you add "hacktoberfest-accepted" label to this PR to make it count. But if for any reason you don't want to do it, I'm also ok with it 👍

akuleshov7:
> One more thing: I did this PR because I'm participating in the [Hacktoberfest](https://hacktoberfest.com/). I would be grateful if you add "hacktoberfest-accepted" label to this PR to make it count. But if for any reason you don't want to do it, I'm also ok with it 👍

Added, am I right?

ILikeYourHat:
> Added, am I right?

Yes, thank you!

ILikeYourHat:
Any update on this PR? Do you need anything more from me?

ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/decoders/PartialDecoderTest.kt
```diff
@@ -1,33 +1,26 @@
...
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlin.test.Test
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.serializer
-import kotlin.test.assertEquals
 
 @ExperimentalSerializationApi
 class PartialDecoderTest {
-    @Serializable
-    data class TwoTomlTables(val table1: Table1, val table2: Table2)
```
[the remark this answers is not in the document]
akuleshov7: I do not see this class anymore in the tests

buildSrc/src/main/kotlin/com/akuleshov7/Versions.kt
```diff
@@ -8,4 +8,5 @@ object Versions {
     const val JUNIT = "5.7.1"
```
akuleshov7, line 8: Why you did not remove Junit then?

akuleshov7: Looks like Junit still remains @ILikeYourHat what was the idea?

ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/decoders/PartialDecoderTest.kt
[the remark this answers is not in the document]
ILikeYourHat: It was removed because it was actually not used.

In this test you are providing a toml model which as a whole is parseable to `TwoTomlTables`. However, the thing that is being tested here is partial decoding. So you don't need to use the whole model, only the model of the part you are decoding. In other words, the `TwoTomlTables` model was used only to get the `table2` value, which seems like a boilerplate to me

buildSrc/src/main/kotlin/com/akuleshov7/Versions.kt
ILikeYourHat: Migrating to Kotest is a big change, and my idea was to migrate step by step, PR by PR, to make CR easier and avoid possible conflicts. However I can migrate all tests at once in this PR, if you prefer that way. It is up to you :)

akuleshov7: If we do a migration - it is difficult to keep both Unit and Kotest :( 
Better to migrate everything, sorry 🙏

ILikeYourHat: @akuleshov7 it took a while, but here you have it: everything migrated to Kotest :)

Please note that the actual runner for jvm is still JUnit5. Kotest also have an jvm test runner, but the test format is very different there: https://kotest.io/docs/framework/writing-tests.html. I know, the issue mentions also the change of the runner, but in my opinion:
- the main benefit are assertions, and those are migrated
- things like parametrized test, user friendly names of tests etc. can be achieved also in JUnit5 runner
- this PR is big enough, and I don't have enough energy for increasing the scope

PR #246, opened 2023-12-03
Initial Support for Simple Map decoding with unnamed toml keys

akuleshov7:
### What's done:

- Phaze 1: only primitive tables (without nested multiple tables) with no explicit  type checking.
- The idea is to decode unnamed arguments.
For example: 
```
[map]
a = 1
b = 2 
```

or 
```
map = {a=1, b=2}
```

Should be decoded (without knowing and naming explicitly toml keys) to:
```
data class Test {
    val map: Map<String, Int>
}
```

Guys like @jakubgwozdz @y9san9 requested it and it looks like it can be useful for parsing gradle configs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,6 +287,10 @@ gradle-libs-like-property = { id = "org.jetbrains.kotlin.jvm", version.ref = "ko
     property1 = 100
     property2 = 6
 
+[myMap]
+    a = "b"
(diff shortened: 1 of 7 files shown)


akuleshov7:
@y9san9 implemented as we have discussed with you 😄

PR #253, opened 2024-01-14
Fix Serialization in readme to use encodeToString instead of wrong decodeFromString

vlsi:


akuleshov7:
Thank you for your contribution, Volodya! Thank you, we really missed that - weird copy-paste issue.

LGTM

PR #270, opened 2024-05-27
Enable linux arm64 compilation

zt64:
I enabled compilation for the linux arm64 target

This required updating a few libraries
Okio -> 3.9.0
Kotlinx datetime -> 0.6.0

I also had to change the linuxX64Main sourceset to account for the new target, so now its just linuxMain


akuleshov7 (approved):

akuleshov7:
lgtm, thank you for this perfect change, I will make a release together with a migration to Kotlin 2

PR #291, opened 2025-03-01
Fix decoding map

### LINKED ISSUE #287: error decoding map: key is not a valid map index
I tried this:

```kotlin
val d = mapOf("key" to "value")
val t = Toml.encodeToString(d)

Toml.decodeFromString<Map<String,String>>(t)
```
It throws an `IllegalArgumentException: key is not a valid map index`.

This seems so simple that maybe I'm using it wrong?

It works fine with `Json` serialization.

BOOMeranGG:
Fixes #287

### DIFF
--- a/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlMapDecoder.kt
+++ b/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlMapDecoder.kt
@@ -3,9 +3,9 @@ package com.akuleshov7.ktoml.decoders
 import com.akuleshov7.ktoml.TomlInputConfig
 import com.akuleshov7.ktoml.exceptions.IllegalTypeException
 import com.akuleshov7.ktoml.exceptions.InternalDecodingException
+import com.akuleshov7.ktoml.tree.nodes.*
 import com.akuleshov7.ktoml.tree.nodes.TomlKeyValue
-import com.akuleshov7.ktoml.tree.nodes.TomlStubEmptyNode
-import com.akuleshov7.ktoml.tree.nodes.TomlTable
+import com.akuleshov7.ktoml.tree.nodes.pairs.keys.TomlKey
 import kotlinx.serialization.DeserializationStrategy
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.descriptors.SerialDescriptor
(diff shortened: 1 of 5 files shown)


ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlMapDecoder.kt
```diff
@@ -99,4 +126,21 @@ public class TomlMapDecoder(
...
     }
+
+    public companion object {
+        /**
+         * @param deserializer - deserializer provided by Kotlin compiler
+         * @param rootNode - root node for decoding (created after parsing)
+         * @param config - decoding configuration for parsing and serialization
+         * @return decoded (deserialized) object of type T
+         */
+        public fun <T> decode(
```
orchestr7, line 137: So what was the problem? You have introduced a new decoder?

BOOMeranGG: Example from test `testNestedRootMapDecoder`, we have this toml:
```
my_key_1 = "my_value_1"
my_key_2 = "my_value_2"
```

And we got Exception `my_key_1 is not a valid map index` in TomlMainDecoder::decodeElementIndex. **rootNode** has 2 TomlKeyValuePrimitive but no TomlTable (ofc, we can't know it on parsing level and anyway we don't have TomlTable.name)

In that case we can:
- Use `TomlMapDecoder.decode()` instead of `TomlMainDecoder.decode()` in `Toml` class after parsing if we have descriptor.kind == StructureKind.MAP
- Do the same in `TomlMainDecoder`
- ? smth else

I decided to make the first option just because it's easier :)

orchestr7: Yeah, sounds very reasonable, LGTM!

orchestr7 (approved):

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlMapDecoder.kt
BOOMeranGG: I don't have strong opinion so far, how do you think we can avoid TomlMainDecoder?

PR #292, opened 2025-03-01
Fix quoted and dotted keys

### LINKED ISSUE #235: Support for quoted keys and values with special symbols
`String.validateSymbols()` in `ktoml-core-jvm:0.5.0` performs an extra `trimQuotes()`, which prevents assignment to quoted keys containing special symbols.

For example, consider the following line of valid (per [language docs](https://github.com/toml-lang/toml/blob/main/toml.md#keys)) TOML:

```
"androidx.activity:activity-compose" = {}
```

This yields the following exception:

```
ParseException: <...>: Not able to parse the key: ["androidx.activity:activity-compose"] as it contains invalid symbols.
```

BOOMeranGG:
Fixes #235

- Removed extra `trimQuotes()` in `String.validateSymbols()` to allow special symbols in quoted keys #235

- Added `trimSingleQuotes()` in `TomlKey.last()` to fix [this](https://github.com/orchestr7/ktoml/issues/235#issuecomment-2081630668) 

- Renamed `String.getCommentStartIndex` to `String.indexOfFirstOutsideQuotes` and extended it's usage to search for any symbol, which allows to use '=' in quoted keys

- Added extra `.trim()` in `String.splitKeyToTokens` to fit the following requirements:
  image

### DIFF
--- a/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/pairs/keys/TomlKey.kt
+++ b/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/pairs/keys/TomlKey.kt
@@ -4,8 +4,7 @@ import com.akuleshov7.ktoml.TomlConfig
 import com.akuleshov7.ktoml.TomlOutputConfig
 import com.akuleshov7.ktoml.exceptions.TomlWritingException
 import com.akuleshov7.ktoml.parsers.splitKeyToTokens
-import com.akuleshov7.ktoml.parsers.trimQuotes
-import com.akuleshov7.ktoml.parsers.trimSingleQuotes
+import com.akuleshov7.ktoml.parsers.trimAllQuotes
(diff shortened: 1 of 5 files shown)


ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/pairs/keys/TomlKey.kt
```diff
@@ -49,7 +50,10 @@ public class TomlKey internal constructor(
      * Gets the last key part, with all whitespace and quotes trimmed, i.e. `c` in
      * `a.b.' c '`
      */
-    public fun last(): String = keyParts.last().trimQuotes().trim()
+    public fun last(): String = keyParts.last()
+        .trimQuotes()
+        .trimSingleQuotes()
```
orchestr7, line 55: May be merge methods with `trimQuotes` if it is possible?

orchestr7 (approved):

PR #293, opened 2025-03-03
Prohibit default values with a special option

### LINKED ISSUE #67: Prohibit the default values for keys with a special option
Create a special option (disabled by the default) that would control the possibility of the default value for the toml key: 

`class A (val a: B = B(value))`

As some people would like to have a strict control to the data - it could be very useful.

BOOMeranGG:
Fixes #67

### DIFF
--- a/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/TomlConfig.kt
+++ b/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/TomlConfig.kt
@@ -80,6 +80,7 @@ public open class TomlConfig(
  * @property allowNullValues Whether to allow/prohibit null values: a = null
  * @property allowEmptyToml Whether empty toml can be processed, if false - will throw an exception
  * @property allowEscapedQuotesInLiteralStrings Whether to allow/prohibit escaping of single quotes in literal strings
+ * @property ignoreDefaultValues Whether to ignore default values
  */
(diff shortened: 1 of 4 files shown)


ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/TomlConfig.kt
```diff
@@ -80,13 +80,15 @@ public open class TomlConfig(
...
  * @property allowEscapedQuotesInLiteralStrings Whether to allow/prohibit escaping of single quotes in literal strings
+ * @property ignoreDefaultValues Whether to ignore default values
  */
 public data class TomlInputConfig(
     public val ignoreUnknownNames: Boolean = false,
     public val allowEmptyValues: Boolean = true,
     public val allowNullValues: Boolean = true,
     public val allowEmptyToml: Boolean = true,
     public val allowEscapedQuotesInLiteralStrings: Boolean = true,
+    public val ignoreDefaultValues: Boolean = false,
```
orchestr7, line 91: Can you please add this option to readme? 🙏

orchestr7: Also, why do we really need to ignore them? What is the use case?

orchestr7 (approved):
After adding the readme please squash commits and merge PR, I think that you should have these rights :)

PR #294, opened 2025-03-06
Support decoding nested arrays and arrays of multiline strings

BOOMeranGG:
- Support decoding nested arrays
- Support decoding arrays of multiline strings
- Refactor multiline values parsing
- Fix incorrect parsing toml with `' " ] [` inside string which is an element of array


orchestr7 (approved):
LGTM!

PR #297, opened 2025-03-16
Support Nested Inline Tables

### LINKED ISSUE #107: Add the proper support of nested inline tables
```
table2 = { someNumber = 5, table2."akuleshov7.com" = { name = 'this is a "literal" string', configurationList = ["a",  "b",  "c", null] }}
```

BOOMeranGG:
Fixes #107

- Support Nested Inline Tables
- Support arrays in inline tables

First problem was in `TomlInlineTable::parseInlineTableValue`, when we called `.split(",")`. It doesn't work with arrays (cause they also have comma) and with nested inline tables. Now we do the same split, but ignore all commas inside arrays/nested inline.

Second problem was in `TomlInlineTable::returnTable`, in test case we had this TomlTable fullTableKey:
image

### DIFF
--- a/ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/decoders/InlineTableDecoderTest.kt
+++ b/ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/decoders/InlineTableDecoderTest.kt
@@ -5,7 +5,9 @@ import com.akuleshov7.ktoml.exceptions.ParseException
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.decodeFromString
-import kotlin.test.*
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
 
 class InlineTableDecoderTest {
     @Serializable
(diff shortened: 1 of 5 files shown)


ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/decoders/InlineTableDecoderTest.kt
```diff
@@ -30,22 +32,78 @@ class InlineTableDecoderTest {
...
+            @SerialName("akuleshov7.com")
+            val inlineTable: NestedTable,
+        )
+
+        @Serializable
+        data class MyClass(
+            val table2: Table2,
+        )
+
+        val test =
```
orchestr7, line 54: Hm, it looks like deserialization should go in a little bit different way: https://github.com/toml-lang/toml/discussions/879

```
table2 = { table2."akuleshov7.com" = { name = "test" }, i = 1 }
```

is equal to the following one:
```
[table2]
    i = 1
[table2.table2."akuleshov7.com"]
    name = "test"
```

orchestr7: So in `Table2` class there should be another one `table2` field

orchestr7: Will be also great if you could add a test which deserializes the following code:
```
[table2]
    i = 1
[table2.table2."akuleshov7.com"]
    name = "test"
```

to same structures

BOOMeranGG: Done

orchestr7 (approved):

ktoml-core/src/commonTest/kotlin/com/akuleshov7/ktoml/decoders/InlineTableDecoderTest.kt
orchestr7: Thanks, brat! LGTM. Free to merge (don't forget to squash)

PR #300, opened 2025-04-03
Fix commented tables

BOOMeranGG:
Fix for `commentedTableTest`. We had an error:
```
expected:

# Comment 1
# Comment 2
[tableA]
    a = ""

[tableB] # Inline comment
    b = 7

actual:

# Comment 1
# Comment 2
[tableA]
    # Comment 1
    # Comment 2
    a = ""

[tableB] # Inline comment
    b = 7 # Inline comment
```

`TomlEncoderAttributes.set` reset comments only if we have @TomlComments, but we need to reset comments anyway.

In other tests I just deleted @Ignore, they've already been fixed


orchestr7 (approved):
LGTM!

orchestr7:
Weird, that we have missed that

PR #301, opened 2025-04-04
Support unsigned integers

### LINKED ISSUE #173: Support for Unsigned Integer values
Would you mind elaborating on why unsigneds can't be supported?

I'd recommend making a new GitHub issue 'support unsigned ints' (I could make the issue if you like) and adding the link in the source code, then it's easier to track and follow progress.

_Originally posted by @aSemy in https://github.com/akuleshov7/ktoml/pull/171#discussion_r1058245318_

BOOMeranGG:
Fixes #173 

> Neither signed, nor unsigned integers are subtypes of each other.

So there's a bit of code duplication in parsing and serialization between unsigned types and integers, but nothing crucial

### DIFF
--- a/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
+++ b/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
@@ -1,15 +1,19 @@
 package com.akuleshov7.ktoml.decoders
 
 import com.akuleshov7.ktoml.exceptions.IllegalTypeException
+import com.akuleshov7.ktoml.exceptions.InternalDecodingException
 import com.akuleshov7.ktoml.tree.nodes.TomlKeyValue
 import com.akuleshov7.ktoml.tree.nodes.pairs.values.TomlBasicString
 import com.akuleshov7.ktoml.tree.nodes.pairs.values.TomlDouble
 import com.akuleshov7.ktoml.tree.nodes.pairs.values.TomlLiteralString
 import com.akuleshov7.ktoml.tree.nodes.pairs.values.TomlLong
+import com.akuleshov7.ktoml.tree.nodes.pairs.values.TomlUnsignedLong
 import com.akuleshov7.ktoml.utils.FloatingPointLimitsEnum
 import com.akuleshov7.ktoml.utils.FloatingPointLimitsEnum.*
 import com.akuleshov7.ktoml.utils.IntegerLimitsEnum
 import com.akuleshov7.ktoml.utils.IntegerLimitsEnum.*
+import com.akuleshov7.ktoml.utils.UnsignedIntegerLimitsEnum
+import com.akuleshov7.ktoml.utils.UnsignedIntegerLimitsEnum.*
 import com.akuleshov7.ktoml.utils.convertSpecialCharacters
 import kotlinx.datetime.Instant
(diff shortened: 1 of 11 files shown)


ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
```diff
@@ -123,6 +143,26 @@ public abstract class TomlAbstractDecoder : AbstractDecoder() {
         }
     }
 
+    private inline fun <reified T> decodeUnsignedPrimitiveType(): T {
+        val keyValue = decodeKeyValue()
+        try {
+            return when (val value = keyValue.value) {
+                is TomlLong -> decodeUnsignedInteger((value.content as Long).toULong(), keyValue.lineNo)
```
orchestr7, line 150: hm, why for TomlLong we have `decodeUnsignedInteger`? It should be signed, no?

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/pairs/TomlKeyValue.kt
```diff
@@ -145,6 +145,7 @@ public fun String.parseValue(lineNo: Int, config: TomlInputConfig): TomlValue =
         // ===== basic strings
         '\"' -> TomlBasicString(this, lineNo)
         else -> tryParseValue<NumberFormatException>(lineNo, ::TomlLong)  // ==== integer values
+            ?: tryParseValue<NumberFormatException>(lineNo, ::TomlUnsignedLong)  // ===== unsigned integer values
```
orchestr7, line 148: It looks like we will miss sometimes here, and I do not know, how to fix this situation. 

For example, we receive: UnsignedLog with value "1". How would we parse it? 

To SignedLong, right? :(

orchestr7 (approved):
LGTM, but please see comments and linter suggestions

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
BOOMeranGG:
```kotlin
 * TomlUnsignedLong - Toml AST Node for a representation of Arbitrary 64-bit unsigned integers: key = 9_223_372_036_854_775_808
 * This node contains integer in range: [2^63, 2^64-1]. Integers less than 2^63 are treated as TomlLong
```
On parsing level we just see the number and don't know it's type

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/pairs/TomlKeyValue.kt
BOOMeranGG: > For example, we receive: UnsignedLong with value "1". How would we parse it?

As TomlLong, but then we can convert it to ULong. But now I see what I missed:
- When we have UByte/UShort/UInt, and toml contains negative number (let's say -1), then we got an error: "**The unsigned integer literal, that you have provided is <18446744073709551615>, but the type for deserialization**"
- When we have ULong and "-1" in toml, then the result is **18446744073709551615**

So I think we just need to add a check on decoding level to prevent this. If we have negative number and Unsigned type - throw an exception

PR #306, opened 2025-04-21
Implement decoding for inline array of tables

### LINKED ISSUE #99: Inline arrays of tables
TOML standard allows inline arrays of tables in the same way how it is done for single inline tables (https://github.com/akuleshov7/ktoml/issues/48)

```
points = [ { x = 1, y = 2, z = 3 },
           { x = 7, y = 8, z = 9 },
           { x = 2, y = 4, z = 8 } ]

```

After https://github.com/akuleshov7/ktoml/issues/88 will be finished - this issue should be implemented

### LINKED ISSUE #259: According to the TOML specification string values
Line 10: According to the TOML specification string values (even Enums) should be wrapped (start and end) with quotes (""), but the following value was not: <{ name = "user">.

image

BOOMeranGG:
Fixes #99 
Fixes #259

### DIFF
--- a/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/parsers/StringUtils.kt
+++ b/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/parsers/StringUtils.kt
@@ -233,6 +233,15 @@ internal fun String.indexOfNextOutsideQuotes(
 }
 
 /**
+ * @param prefix - the string to check for
+ * @return true if the string starts with the prefix, ignoring all whitespaces
+ */
+@Suppress("FUNCTION_BOOLEAN_PREFIX")
+internal fun String.startsWithIgnoreAllWhitespaces(prefix: String): Boolean = this.filter {
+    !it.isWhitespace()
+}.startsWith(prefix)
+
+/**
  * @param allowEscapedQuotesInLiteralStrings value from TomlInputConfig
  * @param placeholder - the string to replace escaped quotes with
  * @return the string with escaped quotes replaced with a placeholder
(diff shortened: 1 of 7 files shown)


ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/parsers/StringUtils.kt
```diff
@@ -232,6 +232,15 @@ internal fun String.indexOfNextOutsideQuotes(
     return -1
 }
 
+/**
+ * @param prefix - the string to check for
+ * @return true if the string starts with the prefix, ignoring all whitespaces
+ */
+@Suppress("FUNCTION_BOOLEAN_PREFIX")
+internal fun String.startsWithIgnoreAllWhitespaces(prefix: String): Boolean = this.filter {
+    !it.isWhitespace()
```
orchestr7, line 241: filterNot { it.isWhitespace() }

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/tables/TomlInlineTable.kt
```diff
@@ -146,11 +162,90 @@ public class TomlInlineTable internal constructor(
...
+                .replaceEscaped(allowEscapedQuotesInLiteralStrings)
+                .removePrefix("[")
+                .removeSuffix("]")
+
+            val result: MutableList<String> = mutableListOf()
+            val current = StringBuilder()
+            val lastIdx = clearedString.length
+            var i = 0
+
+            while (i < lastIdx) {
```
orchestr7, line 210: sorry, for being a nerd, but can you rewrite the code in kotlin style, please 🙏🏾 
can you please use `when` here?

``` 
when (val char = content[i]) {
                '\"', '\'' -> {
```

orchestr7, line 210: how about `forEachIndexed` ? :)

orchestr7 (approved):
lgtm, but pls can you make it more kotlin-elegant? :)

PR #307, opened 2025-04-21
Add validation for unsigned numbers

### LINKED ISSUE #305: Parsing of unsigned vs signed
It looks like we will miss sometimes here, and I do not know, how to fix this situation. 

For example, we receive: UnsignedLog with value "1". How would we parse it? 

To SignedLong, right? :(

_Originally posted by @orchestr7 in https://github.com/orchestr7/ktoml/pull/301#discussion_r2041119377_

BOOMeranGG:
Fixes #305

### DIFF
--- a/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
+++ b/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
@@ -147,7 +147,11 @@ public abstract class TomlAbstractDecoder : AbstractDecoder() {
         val keyValue = decodeKeyValue()
         try {
             return when (val value = keyValue.value) {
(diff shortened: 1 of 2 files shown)


ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlAbstractDecoder.kt
```diff
@@ -147,7 +147,11 @@ public abstract class TomlAbstractDecoder : AbstractDecoder() {
         val keyValue = decodeKeyValue()
         try {
             return when (val value = keyValue.value) {
-                is TomlLong -> decodeUnsignedInteger((value.content as Long).toULong(), keyValue.lineNo)
+                is TomlLong -> {
```
orchestr7, line 150: I first though that exception should be thrown for negative number and TomlUnsignedLong, but then understood the idea. Got it! LGTM.

orchestr7 (approved):

PR #308, opened 2025-04-23
Add specific error for wrong type when array/map is expected

### LINKED ISSUE #65: Add proper, readable error for incorrect arrays
Now if we will use invalid type where array is expected, ktoml will simply raise:
```
tags is not a valid list index
java.lang.IllegalArgumentException: tags is not a valid list index
```

BOOMeranGG:
Fixes #65

### DIFF
--- a/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlMainDecoder.kt
+++ b/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlMainDecoder.kt
@@ -102,6 +102,7 @@ public class TomlMainDecoder(
         }
 
         val currentNode = rootNode.getNeighbourNodes().elementAt(elementIndex)
+        checkDescriptorHasAllowedType(descriptor)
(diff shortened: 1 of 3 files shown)


ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/decoders/TomlMainDecoder.kt
```diff
@@ -139,6 +140,25 @@ public class TomlMainDecoder(
         return currentProperty
     }
 
+    private fun checkDescriptorHasAllowedType(descriptor: SerialDescriptor) {
+        // LIST and MAP use their own decoders. If we decode with descriptor of these types in
+        // TomlMainDecoder, original toml is invalid, i.e. toml:
+        // a = "abc" # string value
+        // But deserializing type is List/Map
+        if (descriptor.kind == StructureKind.LIST) {
```
orchestr7, line 148: when(descriptor.kind) {
   StructureKind.LIST -> 
   StructureKind.MAP ->
   else ->
}

orchestr7 (approved):
lgtm

PR #312, opened 2025-04-27
Breaking changes: finally removing all deprecated methods and classes (including TOMLConfig)

orchestr7:
### What's done:
- fixed all `Deprecated` warnings, including the callee-code


ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/tables/TomlTable.kt
```diff
@@ -138,6 +127,7 @@ public class TomlTable(
 
                 // Primitive pairs have a single newline after, except when a
                 // table follows.
+                // FixMe: need to clarify why this check for instance is always 'true'
                 if (child !is TomlKeyValuePrimitive || children[index + 1] is TomlTable) {
```
BOOMeranGG, line 131: `child` has a smart cast to `TomlArrayOfTablesElement` due to `if (child is TomlArrayOfTablesElement)`

Looks like it's just a mistake, because this whole method was written once and never changed

BOOMeranGG (approved):

ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/tables/TomlTable.kt
orchestr7: yeah, will remove or check once more in next PRs

PR #326, opened 2025-05-22
Support contextual serializers

### LINKED ISSUE #282: KToml deserialization ignores `serializersModule`
Imagine a serializable class `Foo`, that has a `@Contextual` property `bar` of type `Bar`, just like this:
```kt
@Serializable
class Foo(
    @Contextual
    val bar: Bar
)
```
Then imagine whatever serializer for `Bar` called `BarSerializer`. Now if you want to deserialize `Foo`, you need to add the `BarSerializer` to `serializersModule`. Reading from a file, it should look like this
```kt
val module = SerializersModule {
    contextual(BarSerializer)
}

val reader = TomlFileReader(serializersModule = module)

val foo = reader.decodeFromFile<Foo>(serializer(), "somefile.toml")
```
This doesn't work though - an exception[^1] will be raised. I've tried the equivalent in JSON and everything worked as expected.

When I further dug into the source code, I couldn't find any usage of `serializersModule` in the way of `decodeFromFile` and `decodeFromString`, while it is present in the chain of `encodeToString`.

Exact implementation details for my case can be found [on the Kotlin Slack](https://kotlinlang.slack.com/archives/C7A1U5PTM/p1730736306080469?thread_ts=1730658503.564719&cid=C7A1U5PTM)

[^1]: kotlinx.serialization.SerializationException: Serializer for class '`Bar`' is not found.
Please ensure that class is marked as '`@Serializable`' and that the serialization compiler plugin is applied.

BOOMeranGG:
Fixes #282 

No more EmptySerializersModule in decoders

For encoding I just added a test, it was already working


orchestr7 (approved):
Nice refactoring, thank you!

PR #333, opened 2025-06-18
Fixes around comments in/around multiline strings

cedrickcooke:
I had some pre-existing `.toml` files which embedded markdown, and these managed to trip up this parser. I managed to track down what caused the crashes and included some minimal reproductions as test cases. Before the included fixes these tests would fail, but they now pass.

In the below example the parser would not detect the multiline string as closing, since it required a line ending with `"""`.

```toml
a = """
Multiple lines are required to trigger this bug, triple-quotes are not enough
""" # A comment placed here causes a crash
```

In the below example the parser successfully finds the extent of the mutli-line string, but crashes when splitting the key and value. This happened because the pair is first processed via `takeBeforeComment`, which incorrectly found the markdown header inside the multiline string as a comment. This in turn caused the parser to attempt to assign the value `"""\nThis is an example containing "a quote, where I've used an apostrophe inside." Later, I've used another apostrophe.\n\n` , which crashes when `TomlBasicString` attempts to validate the quotes.

I've tracked the root of the issue to `indexOfNextOutsideQuotes`, which did not handle triple-quotes. The triple quote was handled as a sequence of three individual quotes (open, close, open) which could then be closed by a stray quote inside the string. This didn't cause issues in most cases, where quotes are balanced, but I ended up hitting it with the following pattern:

```toml
a = """
This is an example containing "a quote, where I've used an apostrophe inside." Later, I've used another apostrophe.

# This markdown header was detected as an actual comment
"""
```

### DIFF
--- a/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/parsers/StringUtils.kt
+++ b/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/parsers/StringUtils.kt
@@ -7,6 +7,8 @@ package com.akuleshov7.ktoml.parsers
 import com.akuleshov7.ktoml.exceptions.ParseException
 import com.akuleshov7.ktoml.utils.newLineChar
 
+private const val MULTILINE_STRING_QUOTE_LENGTH = 3
+
 /**
  * Splitting dot-separated string to the list of tokens:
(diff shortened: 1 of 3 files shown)


orchestr7:
Hi! Thank you for a fix, looks very reasonable. I’ve approved the pipeline to run tests

orchestr7 (approved):

PR #334, opened 2025-06-21
Fix parsing of dotted keys inside array-of-tables

cedrickcooke:
Previously, the parser would produce the wrong tree structure for use of `dotted.key = "value"` while inside of an `[[arrayOfTables]]`. As an example:

```toml
[[arrayOfTables]]
dotted.key = "value"
```

Before this fix, this is parsed as:

```
 - TomlFile (rootNode)
     - TomlTable ([[arrayOfTables]])
         - TomlArrayOfTablesElement (technical_node)
     - TomlTable ([dotted])
         - TomlKeyValuePrimitive (key="value")
```

After this fix, it is now correctly parsed as:

```
- TomlFile (rootNode)
     - TomlTable ([[arrayOfTables]])
         - TomlArrayOfTablesElement (technical_node)
             - TomlTable ([arrayOfTables.dotted])
                 - TomlKeyValuePrimitive (key="value")
```

This matches the structure that was already working when using a non-array table:

```toml
[table]
dotted.key = "value"
```

```
 - TomlFile (rootNode)
     - TomlTable ([table])
         - TomlTable ([table.dotted])
             - TomlKeyValuePrimitive (key="value")
```

And this time, I remembered to make sure it passes lint 😄

### DIFF
--- a/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/pairs/TomlKeyValue.kt
+++ b/ktoml-core/src/commonMain/kotlin/com/akuleshov7/ktoml/tree/nodes/pairs/TomlKeyValue.kt
@@ -39,7 +39,11 @@ internal interface TomlKeyValue {
         // updating current KeyValue with this key
         this.key = realKeyWithoutDottedPrefix
         // tables should contain fully qualified name, so we need to add parental name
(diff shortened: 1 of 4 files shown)


orchestr7 (approved):
LGTM! Thank you

